PR #5, opened 2025-07-18
Fix sealing bench

ssnd:


fracmanager/sealer_test.go
```diff
@@ -11,112 +12,112 @@ import (
...
-			cm.CreateSortDocsCache(),
-			&frac.Config{},
-		)
-		err := fillActiveFraction(active, &wg)
-		assert.NoError(b, err)
+	b.ReportAllocs()
 
-		wg.Wait()
-		active.GetAllDocuments() // emulate search-pre-sorted LIDs
+	defer profile.Start( // turn on profiling (look for the file fracmanager/cpu.pprof)
```
dkharms: I am not sure that we need this, because we can explicitly specify `-cpuprofile` flag. Profiling add its own overhead so it's better to have this option as opt-in.

fracmanager/sealer_test.go
```diff
@@ -11,112 +12,112 @@ import (
...
+	runSealingBench(b, &frac.Config{SkipSortDocs: true})
+}
+
+func Benchmark_SealingWithSort(b *testing.B) {
+	runSealingBench(b, &frac.Config{})
+}
 
-	cm, stopFn := getCacheMaintainer()
-	defer stopFn()
+func runSealingBench(b *testing.B, cfg *frac.Config) {
```
dkharms, line 105: I think `b.StopTimer()` is missing here since this logic is all about preparation. And we should call `b.StartTimer()` right before the `frac.Seal()` call.

fracmanager/sealer_test.go
eguguchkin: I would not like the stages of initialization and filling of the fraction to be included in the profile

fracmanager/sealer_test.go
eguguchkin: In case of using b.Loop() this happens automatically.

dkharms (approved):

PR #6, opened 2025-07-18
refactor(sealing): separation of the sealing logic from the implementation of the active faction

ssnd:
## Summary by CodeRabbit

* **New Features**
  * New modular sealing pipeline and streaming sealing source with configurable index builder for faster sealed-index creation and startup.

* **Bug Fixes**
  * Improved resource handling and stricter error checks during reads/unpacking; tighter duplicate document-ID handling.

* **Refactor**
  * Reworked sealing/indexing architecture and consolidated sealing/configuration types for cleaner initialization.

* **Tests**
  * Removed legacy sealing tests and added focused unit tests for block-building and sealing components.

* **Chores**
  * Added small utilities (path sync, writer release) and updated sealing configuration structure.

### DIFF
--- a/fracmanager/sealer_test.go
+++ b/fracmanager/sealer_test.go
@@ -108,7 +108,6 @@ func Benchmark_SealingWithSort(b *testing.B) {
 func runSealingBench(b *testing.B, cfg *frac.Config) {
 	cm := NewCacheMaintainer(uint64(units.MiB)*64, uint64(units.MiB)*64, nil)
 	fp := newFractionProvider(cfg, nil, cm, 1, 1)
-	defer fp.Stop()
 
 	dataDir := filepath.Join(b.TempDir(), "BenchmarkSealing")
 	testscommon.RecreateDir(dataDir)


frac/sealed/sealing/index.go
```diff
@@ -0,0 +1,334 @@
...
+		return err
+	}
+	if _, err = ws.Write(prefix); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *IndexSealer) writeBocks(pos int, payloadWriter, headersWriter io.Writer, src Source) error {
```
forshev: `writeBocks` -> `writeBlocks`

frac/common/info.go
```diff
@@ -1,7 +1,8 @@
-package frac
+package common
```
forshev, line 1: nit: what do you think about splitting common package into `info` and `params` packages?

forshev:
thanks for the comments in the code, they are really helpful!

forshev (approved):

frac/common/info.go
dkharms, line 1: We use several different packages to represent bytes in human-readable format:
- `github.com/alecthomas/units`;
- `github.com/c2h5oh/datasize`;

Let's stick to using only one.

frac/active_sealing_source.go
```diff
@@ -0,0 +1,371 @@
...
+	src.info.SealingTime = uint64(src.created.UnixMilli())
+	src.info.BuildDistribution(func(yield func(seq.ID) bool) {
+		for i, mid := range src.mids.vals {
+			if !yield(seq.ID{MID: seq.MID(mid), RID: seq.RID(src.rids.vals[i])}) {
+				return
+			}
+		}
+	})
+}
+
```
dkharms, line 152:
```suggestion
func (src *ActiveSealingSource) prepareInfo() {
	src.info.MetaOnDisk = 0
	src.info.SealingTime = uint64(src.created.UnixMilli())
	src.info.BuildDistribution(src.ids())
}

func (src *ActiveSealingSource) ids() iter.Seq[seq.ID] {
	return func(yield func(seq.ID) bool) {
		for i, mid := range src.mids.vals {
			seqid := seq.ID{
				MID: seq.MID(mid),
				RID: seq.RID(src.rids.vals[i]),
			}

			if !yield(seqid) {
				return
			}
		}
	}
}
```

dkharms: Also in the line `128` (with content `RID: seq.RID(src.rids.vals[i])`) compiler inserts bound checks. Since we can have a lot of seq ids in the slice maybe we can eliminate bound checks?

```go
func (src *ActiveSealingSource) ids() iter.Seq[seq.ID] {
	return func(yield func(seq.ID) bool) {
		if len(src.mids.vals) != len(src.rids.vals) {
			panic("mids and rids must have equal length")
		}

		mids := src.mids.vals
		rids := src.rids.vals

		for i := range mids {
			seqid := seq.ID{
				MID: seq.MID(mids[i]),
				RID: seq.RID(rids[i]), // No bound checks here
			}

			if !yield(seqid) {
				return
			}
		}
	}
}
```

I've ran benchmarks with both versions (with BCE and without) but haven't noticed any improvements, but maybe the reason is our benchmarks. What do you think?

frac/active_sealing_source.go
```diff
@@ -0,0 +1,371 @@
...
+func (src *ActiveSealingSource) Info() *common.Info {
+	return src.info
+}
+
+func (src *ActiveSealingSource) TokenBlocks(blockSize int) iter.Seq[[][]byte] {
+	const uint32Size = int(unsafe.Sizeof(uint32(0)))
```
dkharms: Here `uint32Size` represents size of reserved bytes for token size - is my understanding correct?

So, for example, for token `foo` we reserve 7 bytes (3 bytes for actual token value and 4 bytes for its length):
```
[_][_][_][3][f][o][o]
```

1. Maybe we should rename `uint32Size` to `tokenReservedLength` or something like this?
2. User can specify `max_token_size` via config and we do not validate the maximal value of this option but it's obvious that it cannot be higher than `math.MaxUint32`;
3. Why we are not using [`binary.Uvarint`](https://pkg.go.dev/encoding/binary#Uvarint) for encoding token's length? Seem like we waste space (for example, to encode 1024 in decimal we would need only 2 bytes but majority of tokens can squeeze into 1 bytes).

frac/sealed/token/table_loader.go
```diff
@@ -105,6 +105,7 @@ func (l *TableLoader) loadBlocks() ([]TableBlock, error) {
 	return blocks, nil
 }
 
+// TableBlock represents how token.Table is stored on disk
 type TableBlock struct {
```
dkharms, line 109: nit: I am correct that `TableBlock` represents token table on disk and represents only one block? IMO naming is a little bit misleading because when I see `Block` I expected multiple occurrences of blocks (not just one).

frac/sealed/token/table_entry.go
```diff
@@ -1,9 +1,8 @@
 package token
 
-// TableEntry describes token.Block metadata: what TID and tokens it contains and etc.
-// One token.Block can cover multiple instances of token.TableEntry
+// TableEntry is part of token.Table and points to a fragment of token.Block
 type TableEntry struct {
```
dkharms, line 4: nit: IMO name does not reflect the purpose of this structure.

Something like `token.BlockDescriptor` (or `TokenBlockDescriptor`) will be more self-explanatory.

frac/common/info.go
eguguchkin: I'm considering the "frac/common" package as a temporary solution.
The "frac/active" package will soon appear in the "New Active Faction" project, along with the "frac/sealed" package, and then "frac/common" will simply become "frac" again (and there will still be no cyclic import).

frac/active_sealing_source.go
```diff
@@ -0,0 +1,371 @@
...
+	return func(yield func([]seq.ID, []seq.DocPos) bool) {
+		mids := src.mids.vals
+		rids := src.rids.vals
+
+		ids := make([]seq.ID, 0, blockSize)
+		pos := make([]seq.DocPos, 0, blockSize)
+
+		// first
+		ids = append(ids, seq.ID{MID: seq.MID(mids[0]), RID: seq.RID(rids[0])})
+		pos = append(pos, 0)
```
dkharms, line 219: Is it some kind of system/internal `seq.ID`?

If not its `seq.DocPos` does not have to be equal to `0`.
If yes we should add comment with explanation.

frac/sealed/sealing/blocks_builder.go
```diff
@@ -0,0 +1,302 @@
...
+type lidsSealBlock struct {
+	ext     lidsExt    // LIDs block metadata for registry marking
+	payload lids.Block // LID data payload
+}
+
+// idsSealBlock represents a sealed block containing various identifier types.
+type idsSealBlock struct {
+	mids   seqids.BlockMIDs
+	rids   seqids.BlockRIDs
+	params seqids.BlockParams
```
dkharms, line 43: Why `params` if this field contains `seq.DocPos` of documents?

frac/active_sealing_source.go
eguguchkin: I'm going to update BuildDistribution to expect the slice of MIDs of the parameters, since RIDs aren't used there at all.

So, BuildDistribution might now look like this:

```go
src.info.BuildDistribution(src.mids.vals)
```

eguguchkin: I will do it in this PR

frac/active_sealing_source.go
eguguchkin: We will likely see a difference in the dataset size. However, this refactoring affects the index format, and such tasks involve additional difficulties related to backward compatibility. Therefore, I would collect more similar changes and release them in a new version of the fraction index, likely as part of the compaction project.

But definitely in a different PR

frac/sealed/token/table_loader.go
eguguchkin: In fact, in the current version of sealing, the token table is divided into several blocks. In my implementation, I combined them into a single block, since we always read and use the entire table, so there's no practical need to store it separately.

However, my implementation still allows this block to be split if necessary.

Regarding naming: historically, we also have sections consisting of a single block (for example, the "docs blocks offsets" block and the "info" block). Therefore, the term "block" is used regardless of whether it is the only one in its section or not.

frac/sealed/token/table_entry.go
eguguchkin: Yep, I thought about this. I like 
 - token.Range 
 - tokens.SubRange 
 - token.Span 
 - token.BlockSpan 
 
Or something like this

And I don't mind renaming

frac/active_sealing_source.go
eguguchkin: This is a reserved position (zero) because our LIDs are 1-based. It is part of the index format, and I have not modified this in the PR.

frac/sealed/sealing/blocks_builder.go
eguguchkin: Params it is historical name.

The term "Positions" is excessively long in my opinion. However, its abbreviated form, "Pos," is ambiguous as it does not clearly indicate whether it is singular or plural.

Example:

```go

for _, pos := range pos { // Unclear: single `pos` vs. slice of `pos` (plural)
    // ...
}

```

That's why I'm keeping the name Params :)

frac/active_sealing_source.go
eguguchkin: I actually had a few ideas for improving the data encoding and block header markup, but I'm saving them for when we can no longer delay changing the format.

frac/active_sealing_source.go
eguguchkin: Add comment

frac/active_sealing_source.go
eguguchkin: Done

frac/common/info.go
eguguchkin: Definitely, this should not be done in this already large PR

dkharms: Yep, totally agree. I'll create an issue.

frac/sealed/sealing/index.go
eguguchkin: done

frac/sealed/sealing/blocks_builder.go
```diff
@@ -0,0 +1,306 @@
...
+//   - tokenBatches: Iterator of token batches, where each batch becomes a separate block
+//   - fields: Iterator of [fieldName, maxTID] pairs for all fields in ascending TID order
+//
+// Returns: Iterator of [token block, field table for block] pairs, where field table contains
+// information about which fields and their ranges are represented in this block.
+func (bb *blocksBuilder) BuildTokenBlocks(
+	tokenBatches iter.Seq[[][]byte],
+	fields iter.Seq2[string, uint32],
+) iter.Seq2[tokensSealBlock, []token.FieldTable] {
+	return func(yield func(tokensSealBlock, []token.FieldTable) bool) {
```
dkharms, line 80: You know, I am interested why you made decision to stick with the signature `iter.Seq2[tokensSealBlock, []token.FieldTable]` instead of `iter.Seq2[token.FieldTable, []tokenSealBlock]`?

Like we return one field and how many token blocks it occupies - this just seems more natural and intuitive to me (but this is very subjective).

Also (I believe) signature choice can have impact on performance. Imagine that we have 10k token blocks and only 100 fields which values are distributed among these blocks.

If we stick with `iter.Seq2[token.FieldTable, []tokenSealBlock]` method `BuildTokenBlocks` will yield only 100 times instead of 10k because we have 100 fields to yield (less `call` instruction invocations, less callee/caller registers bookkeeping etc.).

But I do not have any proof for my statement - just a thoughts.

frac/sealed/sealing/blocks_builder.go
```diff
@@ -0,0 +1,306 @@
...
+//   - blockIndex: Index of the current token block
+//   - block: Current token block data
+func createTokenTableEntry(entryStartTID, fieldMaxTID, blockIndex uint32, block tokensSealBlock) *token.TableEntry {
+	// Convert global TIDs to block-local indices
+	firstIndex := entryStartTID - block.ext.minTID
+	lastIndex := min(fieldMaxTID, block.ext.maxTID) - block.ext.minTID
+
+	// Extract min and max token values for the entry range
+	minVal := string(block.payload.GetToken(int(firstIndex)))
+	maxVal := string(block.payload.GetToken(int(lastIndex)))
```
dkharms, line 139: Could we use here `util.ByteToStringUnsafe`?

fracmanager/sealed_frac_cache_test.go
```diff
@@ -13,8 +13,9 @@ import (
 
 	"github.com/ozontech/seq-db/consts"
 	"github.com/ozontech/seq-db/frac"
+	"github.com/ozontech/seq-db/frac/common"
 	"github.com/ozontech/seq-db/seq"
-	"github.com/ozontech/seq-db/tests/common"
+	tests_common "github.com/ozontech/seq-db/tests/common"
```
dkharms:
```suggestion
	testscommon "github.com/ozontech/seq-db/tests/common"
```

Sorry, but nit :shipit:

frac/active_sealing_source.go
```diff
@@ -0,0 +1,460 @@
...
+// sortFields sorts field names and calculates maximum TIDs for each field.
+// Returns sorted field list and array of maximum TIDs.
+func sortFields(tl *TokenList) ([]string, []uint32) {
+	fields := make([]string, 0, len(tl.FieldTIDs))
+	for field := range tl.FieldTIDs {
+		fields = append(fields, field)
+	}
+	slices.Sort(fields)
+
+	pos := 0
```
dkharms, line 113: Somewhere in the PR you said that sequence of TID values starts with 1, so shouldn't be this initial value of `pos` equal to 1?

frac/sealed/sealing/blocks_builder.go
eguguchkin: 1. Regarding the core logic and idea:

- The function's purpose is to transform a sequence of token batches into a sequence of token blocks. In the process, it annotates each block with its sequence number, TID range, and packs the tokens into a special structure consisting of a payload and offsets.
- Each block is returned along with an auxiliary structure—a set of ranges linked to fields.
- The underlying logic of range annotation implies that a single block cannot contain two ranges with the same field (see the diagram).
- However, a block can contain many ranges with different fields. Thus, the relationship from a block to its ranges is 1:N.
- If we return iter.Seq2[token.FieldTable, []tokenSealBlock] (as you suggest), a single token.FieldTable might be associated with only a **fragment of a single block**.
- Furthermore, the segmentation of ranges depends on the segmentation of blocks (the left and right ranges are aligned to block boundaries). This means that, logically, the block is primary and defines the ranges, not the other way around.
- In summary, such a mapping would not work logically. It's simply impossible to link range -> blocks instead of block -> ranges (again, refer to the diagram).

image
 
2. Regarding performance:

- Yes, we could return multiple blocks at once.
- However, a single block already contains thousands of tokens. I believe it's unlikely we would notice any performance difference by returning two, three, etc., blocks at a time.
- On the other hand, doing so would increase memory consumption.

frac/sealed/sealing/blocks_builder.go
eguguchkin: This is impossible. We get a block from the iterator, pack it, and send it on to the next write (where it's copied to the buffer), and on the next iteration, we reuse this structure to fill the new block with data.

But we accumulate MinVal and MaxVal over all iterations, and if we don't copy them from the block, they will become corrupted.

image

frac/active_sealing_source.go
eguguchkin: Let’s consider the case where the first field has only one token. In the first iteration, we will match the field name at the zero position fields[0] to the maximum TID value at the zero position maxTIDs[0],
which will be equal to 1 (specifically: `pos += len(tl.FieldTIDs[field])`).

Overall, this will correspond to reality — for the first field, the maximum TID = 1.
In the general case, the field range is defined as (prevMax, currMax], and in our example, it is (0, 1].

fracmanager/sealed_frac_cache_test.go
eguguchkin: done

frac/active_sealing_source.go
dkharms: Yeah, I got it, thanks!

PR #15, opened 2025-07-18
Seal fractions immediately after reply on start

ssnd:
## Summary by CodeRabbit

* **Refactor**
  * Separated handling of active vs sealed fractions and simplified loading flow.
  * Restored replay-driven initialization of active fractions so only non-empty actives become current; sealed fractions are appended directly.
  * Rotation now uses the manager's internal active-creation helper for consistent initialization.

* **Chores**
  * Simplified method signatures and removed an unused context parameter to reduce complexity.

### DIFF
--- a/fracmanager/fracmanager.go
+++ b/fracmanager/fracmanager.go
@@ -287,8 +287,6 @@ func (fm *FracManager) Start() {
 }
 
 func (fm *FracManager) Load(ctx context.Context) error {
-	var err error
-
 	l := NewLoader(fm.config, fm.fracProvider, fm.fracCache)
 
 	actives, sealed, err := l.load()
@@ -322,9 +320,6 @@ func (fm *FracManager) Load(ctx context.Context) error {
 }
 
 func (fm *FracManager) replayAll(ctx context.Context, actives []*frac.Active) error {
-	wg := sync.WaitGroup{}
-	defer wg.Wait()
-
 	for i, a := range actives {
 		if err := a.Replay(ctx); err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


fracmanager/fracmanager.go
```diff
@@ -280,11 +288,19 @@ func (fm *FracManager) Start() {
 
 func (fm *FracManager) Load(ctx context.Context) error {
 	var err error
```
moflotas: nit: can be removed

fracmanager/fracmanager.go
```diff
@@ -298,16 +314,39 @@ func (fm *FracManager) Load(ctx context.Context) error {
...
+		if i == len(actives)-1 { // last and not empty
+			fm.active = r
+		} else {
+			wg.Wait() // wait previous sealing complete
+
+			wg.Add(1)
+			go func() {
+				fm.seal(r)
+				wg.Done()
+			}()
```
moflotas: nit: maybe reduce nesting?
```suggestion
		if i == len(actives)-1 { // last and not empty
			fm.active = r
			continue
		}
		wg.Wait() // wait previous sealing complete

		wg.Add(1)
		go func() {
			fm.seal(r)
			wg.Done()
		}()
```

moflotas: seems a little overcomplicated, while it is doing basically the same as synchronous sealing

moflotas (changes requested):

fracmanager/fracmanager.go
eguguchkin: Completely got rid of asynchronous sealing.

But we need implement extended async sealing in next part of this task

moflotas (approved):

PR #22, opened 2025-07-18
feat(seqdb): support configuration via file

### LINKED ISSUE #21: Feature: Add configuration via file
Now seq-db is configured via flags. It becomes a little bit cumbersome since we have around 60 flags (and of course there will be more in the future).

So we decided to move configuration options into separate file.

In flags we decided to 
- `--mode` flag;
- `--config` flag;

dkharms:
Closes #21 

## Summary by CodeRabbit

* **New Features**
  * Introduced a unified configuration system with a new config file format, enabling centralized management of application settings.
  * Added an example configuration file for easier setup.

* **Refactor**
  * Migrated most runtime settings from command-line flags to configuration files.
  * Standardized memory size units across the application using a well-known units package.
  * Updated codebase to use the new configuration structure and naming conventions.

* **Chores**
  * Updated dependencies to include configuration and units libraries.
  * Simplified and cleaned up legacy code and removed unused flags.


cmd/seq-db/seq-db.go
```diff
@@ -157,134 +171,138 @@ func startProxy(_ context.Context, addr string, mp bulk.MappingProvider, inMemor
...
+				BucketWidth:              cfg.CircuitBreaker.Bulk.BucketWidth,
+				RequestVolumeThreshold:   int64(cfg.CircuitBreaker.Bulk.VolumeThreshold),
+				ErrorThresholdPercentage: int64(cfg.CircuitBreaker.Bulk.ErrPercentage),
+				SleepWindow:              cfg.CircuitBreaker.Bulk.SleepWindow,
 			},
-			MaxInflightBulks:       *flagMaxInflightBulks,
-			AllowedTimeDrift:       *flagAllowedTimeDrift,
-			FutureAllowedTimeDrift: *flagFutureAllowedTimeDrift,
+			MaxInflightBulks:       cfg.Limits.InflightBulks,
+			AllowedTimeDrift:       cfg.Indexing.AllowedTimeDrift,
```
moflotas: nit: a little bit misleading name, maybe rename it to `PastTimeDrift`?

config/limits.go
```diff
@@ -15,10 +13,14 @@ var (
...
 func init() {
-	_, _ = maxprocs.Set(maxprocs.Logger(func(tpl string, args ...any) { logger.Info(fmt.Sprintf(tpl, args...)) }))
+	_, _ = maxprocs.Set()
 
 	NumCPU = runtime.GOMAXPROCS(0)
 	TotalMemory = getTotalMemory()
+
+	IndexWorkers = NumCPU
+	FetchWorkers = NumCPU
+	ReaderWorkers = NumCPU
```
moflotas, line 23: Maybe put those into config and nearby https://github.com/ozontech/seq-db/pull/22/files#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R23

cmd/seq-db/seq-db.go
dkharms: Done.

config/limits.go
dkharms: I think they are fine where they are now (`limits.go`) since it clear where we can find these limits. And also I wanted to preserve previous file structure.

dkharms:
@coderabbitai review

dkharms:
@coderabbitai pause

Makefile
```diff
@@ -32,11 +32,10 @@ build-image:
 
 .PHONY: run
 run: build-binaries
-	@$(eval DATA_DIR := $(shell mktemp -d))
+	@rm -rdf /tmp/seq-db && mkdir /tmp/seq-db
```
eguguchkin: nit:
I don't really like that we hardcoded the path. There is a mechanism to do this via an environment variable. Let's use that feature here?

As far as I understand, this should work (or something like this):

```
@$(eval DATA_DIR := $(shell mktemp -d))
SEQDB_STORAGE_DATA_DIR=${DATA_DIR} ${LOCAL_BIN}/${OS}-${ARCH}/seq-db
```

dkharms: Yeah, it looks better. Done.

PR #25, opened 2025-07-21
setup codecov report

ssnd:


README.md
```diff
@@ -34,7 +45,7 @@ Join our community for discussions, support, and contributions:
 ## Benchmarks
```
dkharms, line 45: Can you please also capitalize benchmarks section in `docs`?
https://ozontech.github.io/seq-db-docs/seq-db/benchmarks

ssnd: done, I've updated the header in our docs file and will redeploy the docs website later

PR #29, opened 2025-07-23
feat(parser): add ip_range filter

### LINKED ISSUE #28: Feature: ip_range filter
Add ability to search `keyword` fields, which are valid IPs using `ip_range` filter
- `host: ip_range(192.168.1.1, 192.168.255.1)`
- `host: ip_range(192.168.1.1/24)`

moflotas:
Fixes #28 

## Summary by CodeRabbit

- **New Features**
  - Added ip_range filter support (CIDR and start–end ranges) across query parsing, serialization, and search matching, enabling IP-range queries with existing logical operators.

- **Bug Fixes**
  - Standardized parser error messages to "expected …" for clearer feedback.

- **Tests**
  - Expanded test coverage with many ip_range cases and table-driven pattern tests.

- **Chores**
  - Added a networking dependency to support IP parsing and validation.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"cmp"
+	"path/filepath"
 	"time"
 
 	"github.com/alecthomas/units"
@@ -15,9 +16,17 @@ const (
 func Parse(path string) (Config, error) {
 	var c Config
 
+	abs, err := filepath.Abs(path)
+	if err != nil {
+		return Config{}, err
+	}
+
 	if err := fig.Load(
 		&c,
-		fig.File(path),
+		fig.File(filepath.Base(abs)),
+		// To find config file [fig] iterates over directories
+		// and concatenates filepath with each directory.
+		fig.Dirs(filepath.Dir(abs)),
 		fig.UseStrict(),
 		fig.Tag("config"),
 		fig.UseEnv("SEQDB"),


parser/token_ip_range.go
```diff
@@ -0,0 +1,108 @@
...
+	To    Term
+}
+
+func (n *IPRange) Dump(builder *strings.Builder) {
+	builder.WriteString(quoteTokenIfNeeded(n.Field))
+	builder.WriteString(`:ip_range(`)
+
+	n.From.DumpSeqQL(builder)
+	builder.WriteString(", ")
+	n.To.DumpSeqQL(builder)
```
forshev: should it be `n.From.Dump(builder)` and `n.To.Dump(builder)`?

parser/seqql_filter_test.go
```diff
@@ -226,6 +226,17 @@ service:"wms-svc-logistics-megasort" and level:"#"
 	test(`level:"in(one, t,wo)"`, `level:"in(one, t,wo)"`)
 	test(`level:error and k8s_namespace:in(default, kube-system) and k8s_pod:in(kube-proxy-*, kube-apiserver-*, kube-scheduler-*)`,
 		`((level:error and (k8s_namespace:default or k8s_namespace:kube-system)) and ((k8s_pod:kube-proxy-* or k8s_pod:kube-apiserver-*) or k8s_pod:kube-scheduler-*))`)
+
+	// Test filter 'ip_range'.
+	test(`keyword: ip_range(192.168.1.2, 192.168.1.3)`, `keyword:ip_range(192.168.1.2, 192.168.1.3)`)
```
dkharms, line 231: I think we are missing additional tests which verify logical correctness of `ip_range`, not just only whether `ip_range` is lexed and parsed correctly.

parser/token_ip_range.go
```diff
@@ -0,0 +1,108 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		to, err := netip.ParseAddr(tok)
+		if err != nil {
+			return nil, err
+		}
+
+		if from.Compare(to) >= 0 {
```
dkharms, line 90: Are you sure that we should we treat this as a syntactical error?

Yeah, this is tautology since the result set of IPs will always be empty so it makes sense to "fail fast" and return to user this error.

But at the same time in other languages this is perfectly valid AFAIK.

parser/seqql.go
```diff
@@ -215,7 +215,7 @@ again:
 }
 
 func isTokenRune(r rune) bool {
-	return unicode.IsLetter(r) || unicode.IsDigit(r) || r == '_' || r == '.'
+	return unicode.IsLetter(r) || unicode.IsDigit(r) || r == '_' || r == '.' || r == '/'
```
dkharms: Leaving this comment here since `parseCompositeToken` is a caller of `isTokenRune`.

I am not sure that we should handle IP parsing in `parseCompositeToken` function because it allows user to write such queries `a/b/c:foo`. So previously it was an error but now this query is fine.

And for example, if we want to intoduce support for IPv6 it seems like we might break something.

parser/token_ip_range.go
moflotas: I agree that this is not a syntactical error as well as other languages not failing such queries, but I am not sure, whether it is good or not, since it may seem as flaw from our side.

Let's discuss this point with the team

parser/seqql_filter_test.go
moflotas: I added more tests in `pattern` package

parser/seqql.go
moflotas: Changed the logic, so that `parseCompositeToken` can now accept arbitrary runes, that we want to include while parsing the token

parser/token_ip_range.go
dkharms: Well, I am fine with your decision, just was curious about your reasoning.

Issue #38, opened 2025-07-30
Feature: S3 Logs Offloading


dkharms:
## Motivation

Some clients require storing logs for a very long time and being able to search through them. Currently, in production, our logs are stored for about 60 days, which is not much—we would like to have the ability to increase this retention period indefinitely (potentially) and allow users to define the retention period themselves.

## Solution

It is necessary to implement the export of logs to an ice storage (ice storage, because we already have a cold storage).

At the first stage, the ice storage will be S3, so we need to learn how to work with S3. Later, other storage solutions may be added (Clickhouse, NFS, etc.). It should also be possible to search through this data.

PR #58, opened 2025-08-01
refactor: rename `disk` package to `storage`

dkharms:
Since we can have different storage backends (like local FS, S3, etc.), this change will make code easier to maintain in the future.

This is the first commit in the series of commits (#38)

## Summary by CodeRabbit

* **Refactor**
  * Replaced legacy disk-specific backend with a generic storage layer to support multiple storage backends (local/remote) and generic readers.
  * Public types and constructors updated to accept storage abstractions; added storage type indicator and read-only/upload interfaces.

* **Performance / Chores**
  * Sealed-fraction cache now tracks versions to persist only when changed, reducing unnecessary disk writes.

* **Style / Documentation**
  * Improved log messages and comments for clarity (e.g., "frac-cache").

### DIFF
--- a/cmd/distribution/main.go
+++ b/cmd/distribution/main.go
@@ -11,20 +11,20 @@ import (
 
 	"github.com/ozontech/seq-db/cache"
 	"github.com/ozontech/seq-db/consts"
-	"github.com/ozontech/seq-db/disk"
 	"github.com/ozontech/seq-db/frac"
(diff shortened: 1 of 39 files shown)


ssnd:
@allcontributors add @dkharms for code

ssnd (approved):

moflotas (approved):

PR #60, opened 2025-08-01
feat: logs S3 offloading

dkharms:
This commit introduces support for fraction offloading and ability to search, build aggregations from remote (offloaded) fractions.

---

Things to polish:

- [x] Fix all `FIXME` comments;
- [x] Add integration tests;

## Summary by CodeRabbit

- **New Features**
  - Optional S3 offloading for sealed data: remote-backed fractions coexist with local ones and can be offloaded with retention-based cleanup.

- **Configuration**
  - New offloading block (enabled, retention, endpoint, bucket, region, access_key, secret_key); example updated; storage size defaults adjusted (frac_size, total_size).

- **Observability**
  - Offloading metrics added (count, duration); data-size metric now includes storage_type label.

- **Other**
  - Removal helper now tolerates missing files.


frac/sealed.go
```diff
@@ -387,3 +403,41 @@ func (f *Sealed) Contains(id seq.MID) bool {
 func (f *Sealed) IsIntersecting(from, to seq.MID) bool {
 	return f.info.IsIntersecting(from, to)
 }
+
+func loadHeader(
+	baseFileName string,
```
ssnd: is `baseFileName` used anywhere in this function?

frac/remote.go
```diff
@@ -0,0 +1,223 @@
...
+		f.openDocs()
+		f.openIndex()
+
+		(&Loader{}).Load(f.state, f.info, &f.indexReader)
+		f.isLoaded = true
+	}
+}
+
+func (f *Remote) openIndex() {
+	if f.indexFile == nil {
```
ssnd: should we make sure the file exists on the remote storage and skip the fraction initialization in case its' .docs/index files are missing from s3 for some reason?

frac/sealed.go
```diff
@@ -42,19 +43,22 @@ type Sealed struct {
...
 	isLoaded bool
+	state    *State
 
 	readLimiter *storage.ReadLimiter
 
 	// shit for testing
 	PartialSuicideMode PSD
 }
 
+type State struct {
```
ssnd: nit: `SealedState`? active/sealed fraction states mean different things,  the naming should respect that imo.

ssnd: also, is there a  reason  for this struct to stay public?

fracmanager/proxy_frac.go
```diff
@@ -162,6 +159,19 @@ func (f *proxyFrac) trySetSuicided() (*frac.Active, *frac.Sealed, bool) {
 	return active, sealed, sealing
 }
 
+func (f *proxyFrac) Offload(ctx context.Context, u storage.Uploader) (bool, error) {
+	f.useMu.RLock()
+
+	if f.isSealingState() {
```
ssnd, line 165: nit: maybe move useMu locking into f.isSealingState?

fracmanager/sealed_frac_cache_test.go
```diff
@@ -463,8 +461,7 @@ func TestMissingCacheFilesDeleted(t *testing.T) {
 	sealWG2 := sync.WaitGroup{}
 	suicideWG2 := sync.WaitGroup{}
 	fm2.maintenance(&sealWG2, &suicideWG2) // shrinkSizes should be called
-	sealWG2.Wait()
-	suicideWG2.Wait()
+	fm2.Stop()
```
forshev, line 476: i think we should still wait for sealWG and suicideWG

frac/sealed.go
dkharms: No, I forgot to delete this parameter. Will do it.

frac/sealed.go
dkharms: Yep, I'll rename it to `sealedState`. There is no reason to make this struct public.

fracmanager/proxy_frac.go
dkharms: Well, it's done consistently with other methods which use `isSealingState`. Won't fix.

dkharms:
@coderabbitai review

forshev (approved):
lgtm

cmd/seq-db/seq-db.go
```diff
@@ -296,7 +299,28 @@ func startStore(
 			},
 		},
 	}
-	store, err := storeapi.NewStore(ctx, sconfig, mp)
+
+	var s3cli *s3.Client
+	if cfg.Offloading.Enabled {
```
eguguchkin: nit: I would move the creation of the client to a separate function just to make the code easier to read

frac/remote.go
```diff
@@ -0,0 +1,223 @@
...
+func (f *Remote) openDocs() {
+	if f.docsFile == nil {
+		pickedName := path.Base(f.BaseFileName) + consts.DocsFileSuffix
+		sortedName := path.Base(f.BaseFileName) + consts.SdocsFileSuffix
+
+		if ok, _ := f.s3cli.Exists(f.ctx, pickedName); !ok {
+			pickedName = sortedName
+		}
+
+		f.docsFile = s3.NewReader(f.ctx, f.s3cli, pickedName)
```
eguguchkin: nit: it seems that if the file does not exist then we will continue working here and get an error much later when we try to read? I do not claim that this is a big problem but the behavior differs from the behavior of the sealed fraction

fracmanager/fetcher_test.go
```diff
@@ -59,7 +60,7 @@ func testFetcher(t *testing.T, fetcher *Fetcher, hasHint bool) {
 	}
 	ids = append(ids, newID)
 	counter := 0
-	docs, err = fetcher.FetchDocs(context.TODO(), fm.GetAllFracs(), ids)
+	docs, err = fetcher.FetchDocs(context.TODO(), fm.GetFracs(FracTypeLocal), ids)
```
eguguchkin: I don't really like this signature. It seems it would be convenient to continue using the method GetAllFrac()

And it would be suitable here, we don't create remote fractions in the test, so why explicitly limit the type here?

fracmanager/fraction_provider.go
```diff
@@ -66,6 +69,21 @@ func (fp *fractionProvider) NewSealedPreloaded(name string, preloadedData *frac.
 	)
 }
 
+func (fp *fractionProvider) NewRemote(
+	ctx context.Context, name string, cachedInfo *frac.Info, s3cli *s3.Client,
```
eguguchkin: I think that it makes no sense to transfer the client's c3 every time (it doesn't change) but to remember it as a property of the fractionalProvider

fracmanager/proxy_frac.go
```diff
@@ -162,6 +159,19 @@ func (f *proxyFrac) trySetSuicided() (*frac.Active, *frac.Sealed, bool) {
...
+		f.useMu.RUnlock()
+		f.sealWg.Wait()
+		return f.cur().Offload(ctx, u)
```
eguguchkin: Theoretically we could get a suicide state here and cur() would be nil (I know we're either doing an offloading or a suicide now, but while we implement interface we must be ready for this?)

storeapi/grpc_fetch.go
```diff
@@ -68,7 +69,7 @@ func (g *GrpcV1) doFetch(ctx context.Context, req *storeapi.FetchRequest, stream
 	dp := acquireDocFieldsFilter(req.FieldsFilter)
 	defer releaseDocFieldsFilter(dp)
 
-	docsStream := newDocsStream(ctx, ids, g.fetchData.docFetcher, g.fracManager.GetAllFracs())
+	docsStream := newDocsStream(ctx, ids, g.fetchData.docFetcher, g.fracManager.GetFracs(fracmanager.FracTypeLocal|fracmanager.FracTypeRemote))
```
eguguchkin: I agree with the rabbit that it would be more convenient to use the short form GetAllFracs here and in many other places

fracmanager/fracmanager.go
```diff
@@ -130,114 +146,236 @@ func (fm *FracManager) shiftFirstFrac() frac.Fraction {
...
-		}
+	for _, outsider := range outsiders {
+		cleanupWg.Add(1)
+		go func() {
+			defer cleanupWg.Done()
+
+			info := outsider.Info()
+			// Client for offloading is not configured.
+			// So just kill fraction.
+			if fm.s3cli == nil {
```
eguguchkin: I think it's right to check OffloadingEnabled here

fracmanager/fracmanager.go
```diff
@@ -130,114 +146,236 @@ func (fm *FracManager) shiftFirstFrac() frac.Fraction {
...
+			metric.OffloadingDurationSeconds.Observe(float64(time.Since(offloadStart).Seconds()))
+
+			logger.Info(
+				"successully offloaded fraction",
+				zap.String("fraction", info.Name()),
+				zap.String("took", time.Since(offloadStart).String()),
+			)
+
+			// Update storage type for offloaded fraction.
+			info.StorageType = storage.TypeRemote
```
eguguchkin: I didn't find anywhere where we read this field - only write the value

fracmanager/fracmanager.go
```diff
@@ -130,114 +146,236 @@ func (fm *FracManager) shiftFirstFrac() frac.Fraction {
...
+			remote := fm.fracProvider.NewRemote(fm.ctx, info.Path, info, fm.s3cli)
+
+			fm.fracMu.Lock()
+			fm.remoteFracs = append(fm.remoteFracs, &fracRef{remote})
```
eguguchkin: consistency is violated: here a fraction temporarily disappears from the search (fm.localFracs) and then returns in a fm.remoteFracs. So search result will be unstable

fracmanager/fracmanager.go
```diff
@@ -31,15 +32,17 @@ const (
...
 	cacheMaintainer *CacheMaintainer
 
 	fracCache *sealedFracCache
 
-	fracMu sync.RWMutex
-	fracs  []*fracRef
-	active activeRef
+	fracMu      sync.RWMutex
+	localFracs  []*fracRef
+	remoteFracs []*fracRef
```
eguguchkin: fracRef is needed only to replace an instance of the active fraction with a sealed one in any element of the slice

this is not needed in remotes - you can directly store a slice of remotes and avoid unnecessary resolution

frac/fraction.go
```diff
@@ -20,6 +21,7 @@ type Fraction interface {
 	IsIntersecting(from seq.MID, to seq.MID) bool
 	Contains(mid seq.MID) bool
 	DataProvider(context.Context) (DataProvider, func())
+	Offload(ctx context.Context, u storage.Uploader) (bool, error)
```
eguguchkin, line 24: for me this interface is about access to data and not about the life cycle and the offload method seems unnecessary here, especially since out of three implementations of the interface it is implemented in only one - this already indicates an incorrect design

frac/remote.go
```diff
@@ -0,0 +1,223 @@
...
+	docsCache  *cache.Cache[[]byte]
+	docsReader storage.DocsReader
+
+	indexFile   storage.ImmutableFile
+	indexCache  *IndexCache
+	indexReader storage.IndexReader
+
+	loadMu   *sync.RWMutex
+	isLoaded bool
+	state    *State
```
eguguchkin: nit: why pointer here?

frac/sealed.go
```diff
@@ -216,11 +202,41 @@ func (f *Sealed) load() {
...
+		f.openIndex()
+		return u.Upload(gctx, f.indexFile)
+	})
+
+	if err := g.Wait(); err != nil {
+		return true, err
+	}
+
+	remoteFracName := f.BaseFileName + consts.RemoteFractionSuffix
+	if _, err := os.Create(remoteFracName); err != nil {
```
eguguchkin: We need to Close() for file and Sync() for the parent directory here

fracmanager/fracmanager.go
```diff
@@ -138,114 +154,236 @@ func (fm *FracManager) shiftFirstFrac() frac.Fraction {
...
+
+// GetFracs returns a list of known fracs (local or remote).
+//
+// While working with this list, it may become irrelevant (factions may, for example, be deleted).
+// This is a valid situation, because access to the data of these factions (search and fetch) occurs under blocking (see DataProvider).
 // This way we avoid the race.
+//
 // Accessing the deleted faction data just will return an empty result.
-func (fm *FracManager) GetAllFracs() List {
+func (fm *FracManager) GetFracs(ft FracType) (fracs List) {
```
eguguchkin: I would use three methods GetAll - public for searching and getLocal / getRemote - private, they are needed in different cases and so IMHO it is clearer when only the GetAll method is visible to the outside

eguguchkin: in tests it is better to use GetAll()

fracmanager/loader.go
```diff
@@ -153,6 +181,11 @@ func (l *loader) filterInfos(fracIDs []string, infos map[string]*fracInfo) []*fr
 			logger.Panic("frac loader has gone crazy")
 		}
 
+		if info.hasRemote {
```
eguguchkin, line 180: I agree with the rabbit that we are worsening the already bad implementation of cleaning/restoring the state after a failure.

Probably, for this, it is enough to move the PR after checking hasDocsDel

But in general, it will be necessary to separately review this logic of supporting valid states of fractions

frac/fraction.go
dkharms: > interface is about access to data and not about the life cycle

This is a little bit contradictory since we already have `Suicide()` in the interface.
But yeah I'm also not quite good with this method in the interface.

fracmanager/fracmanager.go
dkharms: Yeah, I am aware of this problem but I do not know how to solve it with current implementation of fractions management without blowing up this PR with even bigger diff.

So I see two options:
1. I leave `FIXME` comment and create an issue where we can come up with the new approach for fractions management (and we make it top priority);
1. Or we are trying to fix right here and right now (but hopefully together since I am running out of ideas);

I think first option is OK since offloading is in so-called "alpha" stage.

cmd/seq-db/seq-db.go
dkharms: Yeah, sounds good - I'll make it.

frac/fraction.go
dkharms: Since you've already mentioned it maybe we can come up with new approach.

So you suggested to have lists each of which holds fractions of specific type:
```go
type FracManager struct {
	/* ... */
	activeFracs []*frac.Active
	sealedFracs []*frac.Sealed
	remoteFracs []*frac.Remote
	/* ... */
}
```

Then we can move `Suicide()` and `Offload()` methods from fraction interface to `FracManager` methods e.g.:
```go
func (fm *FracManager) killActive(f *frac.Active) { f.Suicide() }
func (fm *FracManager) killSealed(f *frac.Sealed) { f.Suicide() }
func (fm *FracManager) killRemote(f *frac.Remote) { f.Suicide() }
```

And new `Fraction` interface will look like this:
```go
type Fraction interface {
	Info() *Info
	IsIntersecting(from seq.MID, to seq.MID) bool
	Contains(mid seq.MID) bool
	DataProvider(context.Context) (DataProvider, func())
}
```

After this changes your statement will be truly correct:
> for me this interface is about access to data and not about the life cycle


Yeah so these changes break "polymorphism" but maybe manager has a right to understand with which fraction types its working? And at the same time any other part of codebase will receive fractions via `GetFracs() []frac.Fraction` method.

fracmanager/fracmanager.go
dkharms: I've fixed it in the next [PR](https://github.com/ozontech/seq-db/pull/77/files#diff-0d346fd457fd027447fbc1215f29a16009e631a2afeeb3c9d18746536a4fca45R254) but I'll move this change into this one.

frac/sealed.go
dkharms: Done.

frac/sealed.go
dkharms: Done.

cmd/seq-db/seq-db.go
dkharms: Done.

fracmanager/fetcher_test.go
dkharms: Done.

fracmanager/fraction_provider.go
dkharms: Done.

storeapi/grpc_fetch.go
dkharms: Done.

fracmanager/fracmanager.go
dkharms: I just found it's useful to be able to filter `.frac-cache` like this:
```bash
cat .frac-cache | jq 'select(.storage_type == "remote")'
```

But I am totally fine with removing this property.

frac/remote.go
dkharms: Removed pointer.

fracmanager/fracmanager.go
dkharms: Done.

fracmanager/loader.go
dkharms: Yeah, I agree that loading is messy right now but it's not scope of this PR. Let's fix it later.

fracmanager/fracmanager.go
dkharms: Done.

frac/sealed.go
dkharms: Yeah, missed it - thanks for bringing this up. Done.

fracmanager/sealed_frac_cache_test.go
dkharms: Done.

fracmanager/fracmanager.go
eguguchkin: > I think first option is OK since offloading is in so-called "alpha" stage.

it is ok for me to leave FIXME comment

dkharms: Added `FIXME` comment.

frac/fraction.go
dkharms: We had discussed this problem with @eguguchkin and decided to stick with current approach which will be fixed in the nearest future.

frac/remote.go
dkharms: Fixed it in commit [`93c1423` (#60)](https://github.com/ozontech/seq-db/pull/60/commits/93c14234cac58c7b6a2305374618cefdd068a199) - if some of mandatory files are missing (or S3 experiencing availability issues) will return empty data provider instead of failing. Decided to stick with graceful degradation instead of `exit(1)`.

frac/remote.go
dkharms: Almost identical question was asked by @ssnd.
Answered it here - https://github.com/ozontech/seq-db/pull/60#discussion_r2298417237

fracmanager/fracmanager.go
dkharms: Removed it.

fracmanager/proxy_frac.go
dkharms: Added following check, I think it must be enough:
```go
if c := f.cur(); c != nil {
	return c.Offload(ctx, u)
}
```

fracmanager/fracmanager.go
```diff
@@ -138,114 +154,245 @@ func (fm *FracManager) shiftFirstFrac() frac.Fraction {
...
+				zap.Time("creation_time", ct),
+				zap.String("retention", retention.String()),
+			)
+
+			fm.fracCache.RemoveFraction(f.Info().Name())
+			f.Suicide()
+		}
+	}()
+}
+
```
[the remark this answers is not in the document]
dkharms: Bro, you are delusional.

dkharms:
@coderabbitai pause

fracmanager/proxy_frac.go
eguguchkin: In general, this is not enough. But I expect that we will remove the Offloading method and the problem will be solved in that task

PR #61, opened 2025-08-01
docs: add architecture docs

ssnd:
## Summary by CodeRabbit

* Documentation
  * Added cluster-mode architecture guide describing store/proxy roles, deployment, read/write semantics, replication behavior, and sequence diagrams.
  * Corrected terminology from “faction” to “fraction” across English and Russian docs.
  * Clarified --skip-fsync flag behavior in Store Mode documentation.

* Chores
  * Fixed terminology typos in internal comments and docs (no functional changes).


docs/ru/internal/fractions.md
```diff
@@ -30,10 +30,10 @@ Regular sealing also occurs when the active fraction exceeds the maximum size. T
 
 The sealed fraction consists of 2 files:
 
-* Document file. The same as for the active faction.
-* Index file. This is a serialized representation of the index of the active faction.
+* Document file. The same as for the active .
+* Index file. This is a serialized representation of the index of the active .
 
-Sealing is the process of serializing the in-memory index of the active faction into an index file. 
+Sealing is the process of serializing the in-memory index of the active  into an index file.
```
forshev: in this file `faction` is replaced with `` - empty string

docs/en/12-architecture.md
```diff
@@ -0,0 +1,162 @@
...
+    
+
+  Client->>Proxy: req2
+  Proxy->>B: req2
+  Proxy->>C: req2
+
+  C-->>Proxy: res2_C
+  D-->>Proxy: res2_D
+
+  Proxy-->>Client: merge(res2_C, res2_D)
```
forshev: in the req2 case request goes to replicas B and C but the response is coming from C and D

also i find C and D letters in the response a little misleading, because replicas called `shard2 replica A` and `shard2 replica B`. Maybe change it to something like `res2_2A` and `res2_2B` or rename replicas like `shard2 replica B` and `shard2 replica C`?

forshev (approved):

docs/en/12-architecture.md
```diff
@@ -0,0 +1,162 @@
+# Cluster-Mode Architecture
+
+## Components overview
+
+In cluster mode, seq-db consists of two main components: seq-db store and seq-db proxy
```
dkharms: Let's explicitly clarify that seq-db is just one binary and mode can be controlled via `--mode` flag:

> In cluster mode, seq-db consists of two main components: seq-db store (seq-db instance running with `--mode=store` flag) and seq-db proxy (seq-db instance running with `--mode=proxy` flag).

docs/en/12-architecture.md
```diff
@@ -0,0 +1,162 @@
...
+| `.docs`   | Stores compressed batches of raw log documents |
+| `.meta`   | Tokenized metadata stream (used for recovery)  |
+| `.index`  | On-disk inverted index                         | 
+
+
+Because the dataset is stored in these three file types, moving or restoring a
+shard is straightforward: simply `cp` / `rsync` the directory
+to the target node and start the pod.
+
+Read more about file types and their internal structure [here](./internal/fractions.md).
```
dkharms, line 35: I have not found anything about files with extension `_index`, `docs.del` etc
Should we make a warning that file types in the table are not fully presented?

docs/en/12-architecture.md
```diff
@@ -0,0 +1,162 @@
...
+
+```
+write, fsync   # .meta file
+write, fsync   # .data file
+```
+That is, two write system calls followed by two fsync
+calls—guaranteeing the data survives a node 
+crash or restart before the client receives a success response. 
+Indexing occurs asynchronously, so the newly written
+data becomes searchable shortly after the write is confirmed.
```
dkharms: I think we should give some numbers on how long it takes to make new data visible to search queries. 'shortly' is just so relative.

And I think we should replace 'searchable' with 'visible to search queries`.

docs/en/12-architecture.md
```diff
@@ -0,0 +1,162 @@
...
+traffic distribution without changes to the stateful components
+
+
+#### Key characteristics
+- Deployed as k8s `Deployment`
+- Performs logical replication between stores
+- Routes traffic between storage tiers (hot/cold stores)
+
+seq-db proxy tokenizes every incoming document (new 
+fields become searchable immediately)
```
dkharms: > new fields become searchable immediately

But [here](https://github.com/ozontech/seq-db/pull/61/files#diff-5165ec070f349f71bdc7661e6df6d39e54b7ba44a8052387f4ca1753a2e699e3R47) you said the opposite thing.

docs/en/12-architecture.md
```diff
@@ -0,0 +1,162 @@
...
+  D-->>Proxy: res2_D
+
+  Proxy-->>Client: merge(res2_C, res2_D)
+```
+
+## Notes about replication & consistency
+seq-db doesn't have any mechanism to keep replicas consistent between each other. 
+That is, if a write operation succeeds on a replica of a shard and fails on another replica, the replicas 
+would be out of sync and won't be (automatically) synced. 
+The only given guarantee  is that a write operation will succeed only having at least rf replicas saved on disk.
```
dkharms:
```suggestion
The only given guarantee  is that a write operation will succeed only having at least RF replicas saved on disk.
```

docs/en/12-architecture.md
```diff
@@ -0,0 +1,162 @@
...
+This optimization allows seq-db to have a higher than alternatives ingestion throughput 
+with the obvious price of the possible inconsistencies of retrieval and aggregation queries. 
+seq-db was designed as a database for logs/traces with this tradeoff in mind. 
+It cannot be reliably used for financial calculations or any other operations 
+expected to have high-precision.
```
dkharms: High precision? Maybe something related to consistency?

docs/ru/internal/fractions.md
ssnd: done, thanks

docs/en/12-architecture.md
ssnd: i tried to fix the naming, could you take a look now?

docs/en/12-architecture.md
ssnd: done , thanks!

docs/en/12-architecture.md
ssnd: i've decided to delete this sentence entirely

docs/en/12-architecture.md
ssnd: done, thanks

docs/en/12-architecture.md
ssnd: i've deleted the sentence entirely

docs/ru/internal/fractions.md
```diff
@@ -30,10 +30,10 @@ Regular sealing also occurs when the active fraction exceeds the maximum size. T
 
 The sealed fraction consists of 2 files:
 
-* Document file. The same as for the active faction.
-* Index file. This is a serialized representation of the index of the active faction.
+* Document file. The same as for the active .
+* Index file. This is a serialized representation of the index of the active .
```
forshev: still missing `fraction` word on lines 33 and 34

PR #70, opened 2025-08-04
fix(config): add searchable dir

dkharms:
## Summary by CodeRabbit

* **Chores**
  * Improved configuration file loading to better locate config files by searching within specified directories.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"cmp"
+	"path/filepath"
 	"time"
 
 	"github.com/alecthomas/units"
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -18,6 +18,9 @@ func Parse(path string) (Config, error) {
 	if err := fig.Load(
 		&c,
 		fig.File(path),
+		// To find config file [fig] iterates over directories
+		// and concatenates filepath with each directory.
+		fig.Dirs("/", "."),
```
dkharms: Or I think we should add path to dir that contains config file as separate flag.
Because we can have problems when user passed `--config config.yaml` when running in `seq-db` directory. Example below:

```
- /config.yaml
- /seq-db/config.yaml
```

dkharms:
@coderabbitai summary

moflotas (approved):

PR #77, opened 2025-08-14
test: tests for offloading

dkharms:
Final PR that adds suite for offloading testing.
Definitely it's far from perfect but I have some ideas for future follow-ups.

## Summary by CodeRabbit

- New Features
  - Added ability to force offloading in test environments.
  - Improved readability with string representations for remote and proxy fractions.
- Tests
  - Introduced comprehensive offloading integration suite.
  - Added end-to-end S3/MinIO reader tests and range formatting checks.
- Chores
  - CI now provisions a MinIO service; docker-compose service added for local tests.
  - Makefile gains a target to start test dependencies.
  - YAML schema hints added to workflows.
- Refactor
  - Internal S3 range helper centralized with input validation.
- Dependencies
  - Multiple library versions updated for stability and compatibility.


tests/integration_tests/offloading_test.go
```diff
@@ -0,0 +1,174 @@
+package integration_tests
```
forshev, line 1: nit: maybe this is out of scope of this PR, but i would like to use `integration` build tag for tests that require minio or any other external dependency

storage/s3/reader_test.go
```diff
@@ -0,0 +1,160 @@
...
+		_, err = reader.Read(readData)
+		require.NoError(t, err)
+
+		assert.Equal(t, data[len(data)/2:], readData)
+	})
+
+	t.Run("seek-from-current-and-read", func(t *testing.T) {
+		reader := NewReader(t.Context(), s3cli, path.Base(f.Name()))
+
+		_, err := reader.Seek(int64(len(data)/2), io.SeekCurrent)
```
forshev: nit: maybe move offset firsthand to make current position non-zero?

for example, move offset `reader.Seek(int64(len(data)/2), io.SeekStart)` and then move one more time `reader.Seek(int64((len(data)/2)/2), io.SeekCurrent)` to read the last quarter of a file

tests/integration_tests/offloading_test.go
dkharms: Yeah, definitely we need to do that. And I think not just for external dependencies but also for all integrations test.

storage/s3/reader_test.go
dkharms: Yep, nice suggestion - done!

ssnd (approved):

PR #80, opened 2025-08-15
docs(async search): add async search docs

### LINKED ISSUE #79: Documentation: async searches
Write async searches documentation

forshev:
# Description

Add async search documentation

Fixes #79


docs/en/10-public-api.md
```diff
@@ -684,4 +684,269 @@ Example successful response:
...
+```json
+{
+  "search_id": "c28c97d1-117a-45dc-a0cf-d080f22b2a10"
+}
+```
+
+### `/FetchAsyncSearchResult`
+
+Fetch async search result.
+Returns execution result based on a search parameters from [`/StartAsyncSearch`](#startasyncsearch) request
```
ssnd: maybe we should mention that an this request can return a partial response, depending on whether the search is done or not. and also mention the status field

PR #108, opened 2025-09-02
docs: udpate quickstart to use config files, add seq-ui quickstart

forshev:
# Description

Updated quickstart to use config files
Added seq-ui launch example to quickstart
Updated docker-compose.yaml to use config files

---


config.example.yaml
```diff
@@ -18,7 +22,7 @@ storage:
 offloading:
   enabled: true
   retention: 5m
-  endpoint: http://localhost:9000/
```
dkharms: This endpoint is correct.

config.example.yaml
dkharms, line 1: This file is not intended for use in containers.
It it used in `make run`.

I think for everything related to quickstart we must create directory like `quickstart` or something and there will be `docker-compose.yaml` and other necessary configs.

config.example.yaml
forshev: made a separate config file for docker compose launch

config.example.yaml
forshev: made a separate config file for docker compose launch

.github/workflows/cd.yml
```diff
@@ -41,4 +41,6 @@ jobs:
           build-args: |
             VERSION=${{ github.ref_name }}
             APP_IMAGE=${{ matrix.image.name }}
-          tags: ghcr.io/${{ github.repository_owner }}/${{ github.event.repository.name }}:${{ github.ref_name }}${{ matrix.image.suffix }}
+          tags: |
+            ghcr.io/${{ github.repository_owner }}/${{ github.event.repository.name }}:${{ github.ref_name }}${{ matrix.image.suffix }}
+            ghcr.io/${{ github.repository_owner }}/${{ github.event.repository.name }}:latest${{ matrix.image.suffix }}
```
dkharms, line 46: Do we need here `${{ matrix.image.suffix }}` after `latest` tag?

forshev: this is not used now, but if we decide to build several images based on different base images, we will need it

dkharms (approved):

PR #111, opened 2025-09-03
chore: add config validations

### LINKED ISSUE #85: Add config validations
For now we do not have any config validations - we need to fix it.
Keep in mind that validation depends on `mode` (e.g. `proxy` does not need `total_size`).

dkharms:
# Description

Added validation for config values.

Fixes #85

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

## Summary by CodeRabbit

- New Features
  - Added startup configuration validation that fails fast on invalid settings.
  - Mode-aware checks for proxy, store, and single deployments enforce required fields and numeric ranges.
  - Clear, user-facing errors to help correct misconfigurations before runtime.

- Chores
  - Bumped a concurrency utility dependency to a newer minor version for improved stability.

### DIFF
--- a/config/validation.go
+++ b/config/validation.go
@@ -15,12 +15,12 @@ func (c *Config) Validate(mode string) error {
 
 	switch mode {
 	case "store":
-		validations = append(validations, c.validateStore()...)
+		validations = append(validations, c.storeValidations()...)
 	case "proxy":
-		validations = append(validations, c.validateProxy()...)
+		validations = append(validations, c.proxyValidations()...)
 	case "single":
-		validations = append(validations, c.validateProxy()...)
-		validations = append(validations, c.validateStore()...)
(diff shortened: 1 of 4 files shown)


config/validation.go
```diff
@@ -0,0 +1,114 @@
...
+	for _, fn := range validations {
+		if err := fn(); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (c *Config) validateProxy() []validateFn {
```
moflotas: nit: I suggest adding validation that `hot_read_stores` should always be a subset of `hot_stores`

dkharms: Not sure that this will be useful. I think there might be scenarios where (for whatever reason) I want to have store that is read-only.

moflotas: Yeah, but I believe, you never want having some store write-only (without reading)

PR #145, opened 2025-09-21
fix: fixed panic during tests

### LINKED ISSUE #37: TOCTOU Bug in Circuit Breaker
In each test we create circuit breaker:
```go
func New(name string, config Config) *CircuitBreaker {
	breaker := manager.GetCircuit(name)
	if breaker != nil {
		return &CircuitBreaker{
			Circuit: breaker,
		}
	}
	/* ... */
	breaker = manager.MustCreateCircuit(name, /* ... */)
}
```

Here we have a TOCTOU bug, which sometimes causes test to flap:
```
--- FAIL: TestColdStoreIntegration (21.99s)
    --- FAIL: TestColdStoreIntegration/TestAgg (0.01s)
        manager.go:69: test panicked: circuit with that name already exists
```

romanchechyotkin:
# Description

Added custom suffix for circuit breaker in order not to catch panic via existing name 

Fixes #37

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

## Summary by CodeRabbit

- New Features
  - Added configurable suffix for circuit breaker names, enabling distinct instances per environment or deployment.
- Tests
  - Updated bulk and integration tests to use unique environment and circuit names per test, improving isolation and reducing interference in parallel runs.
  - Enhanced single-suite tests to generate per-environment unique names.
- Chores
  - Standardized test setup to propagate environment-specific suffixes into circuit breaker configuration for consistency across suites.


tests/setup/env.go
```diff
@@ -350,6 +350,7 @@ func MakeIngestors(cfg *TestingEnvConfig, hot, cold [][]string) []*Ingestor {
 					BulkCircuit: circuitbreaker.Config{
 						RequestVolumeThreshold: 101, // disable circuit breaker
 						Timeout:                time.Hour,
+						CustomSuffix:           cfg.Name,
```
dkharms, line 353: Could you please verify that this will fix the issue?

I see that we have several suites (for example `TestBasicIntegration` with name `Basic`) with multiple tests inside those suites. Each test calls `setup.NewTestingEnv` with this `CustomSuffix` so we still can face panics.

Or am I missing something?

romanchechyotkin: yea, you are right, i missed this point

romanchechyotkin: @dkharms already updated test configs, check please

romanchechyotkin: @dkharms friendly reminder

dkharms: @romanchechyotkin sorry, I was chilling on a vacation. Feel free to ping other maintainers.

Honestly, I think your approach is too verbose. I would be fine with something like this (where new CB is created):
```go
breaker := circuitbreaker.New(
	fmt.Sprintf("%s-shard-%d-%s", breakerPrefix, i, uuid.NewString()),
	config,
)
```

Of course, this is a workaround and the real problem is that CB is a global instance (and compiled only once for integration tests).

PR #147, opened 2025-09-24
docs: fix quickstart in 01-quickstart.md

cheb0:
# Description

Currently, the simplest quickstart doesn't work, since `/seq-db-data` doesn't exist inside the container. We can either:
-  simply omit the config for the data directory, since it's the first what a potential user will do and having a simpler config might be a good idea
- create volume and map a local directory to `/seq-db-data`:
```
docker run --rm \
  -p 9002:9002 \
  -p 9004:9004 \
  -p 9200:9200 \
  -v "$(pwd)"/config.yaml:/seq-db/config.yaml \
  -v "$(pwd)"/data:/seq-db-data \
  -it ghcr.io/ozontech/seq-db:latest --mode single --config=config.yaml
```

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

## Summary by CodeRabbit

* **Documentation**
  * Updated Quickstart guides (EN and RU) to simplify the sample config.yaml: removed the storage.data_dir line from the storage block; mapping.path: auto remains.
  * Aligns examples with current defaults to reduce setup confusion and streamline initial configuration.
  * Clarifies that users don’t need to specify data_dir in the basic example.
  * No functional changes to the product; existing configurations are unaffected.

### DIFF
--- a/docs/en/01-quickstart.md
+++ b/docs/en/01-quickstart.md
@@ -22,9 +22,6 @@ Before launch you need to create a config file:
 config.yaml:
 
 ```yaml
-storage:
-  data_dir: /seq-db-data
-
 mapping:
   path: auto
 ```
--- a/docs/ru/01-quickstart.md
+++ b/docs/ru/01-quickstart.md
@@ -22,9 +22,6 @@ seq-db можно быстро запустить в docker-контейнере
 config.yaml:
 
 ```yaml
-storage:
-  data_dir: /seq-db-data
-
 mapping:
   path: auto
 ```


docs/en/01-quickstart.md
```diff
@@ -22,9 +22,6 @@ Before launch you need to create a config file:
 config.yaml:
 
 ```yaml
-storage:
```
dkharms, line 25: Not sure that is related (maybe I'll do that in the #143) but what do you think about adding `make run` in the `Quickstart` section in `README.md`?

From my experience I would like to run freshly compiled binary from `main` branch rather than spin up a container.

forshev (approved):

docs/en/01-quickstart.md
cheb0: That's a good idea, compilation instructions in the quickstart doc (or readme) is also a must.

I also agree with running a binary is a preferred way for us, but for most databases running a container is a way to go, simple because compiling requires a necessary tooling and it might take a long time. So, I assume most people would like to run a container first and make a smoke check.

PR #159, opened 2025-09-26
chore: remove unused code

moflotas:

```
Model: {{model-name}}
Prompt: {{prompt}}
```


proxyapi/ingestor.go
```diff
@@ -47,38 +47,37 @@ type Ingestor struct {
...
+func appendClients(clients map[string]storeapi.StoreApiClient, shards [][]string) error {
 	for _, shard := range shards {
 		for _, replica := range shard {
 			if _, has := clients[replica]; has {
 				continue
 			}
 			// this doesn't block, and if store is down, it will try to reconnect in background
-			conn, err := grpc.DialContext(
-				ctx,
+			conn, err := grpc.NewClient(
```
dkharms, line 80: There is a comment for `DialContext` method:
```
// One subtle difference between NewClient and Dial and DialContext is that the
// former uses "dns" as the default name resolver, while the latter use
// "passthrough" for backward compatibility.  This distinction should not matter
// to most users, but could matter to legacy users that specify a custom dialer
// and expect it to receive the target string directly.
```

I believe these is no difference for us, but could please you double-check it? (if you have not done this already, of course)

PR #168, opened 2025-09-30
perf(search): reduce the influence of the stopwatch

### LINKED ISSUE #158: Add an alternative API to the Stopwatch package for high-performance micro-measurements within loops
The current implementation exhibits slowdowns due to map lookups when accessing metric elements inside loops. In loops with a high number of iterations (on the order of a million), this becomes apparent in the profiler.

It is necessary to extend the API with the ability to pre-allocate metrics outside the loop. This will allow obtaining a direct reference to the metric object, without a map lookup on each iteration, which should reduce CPU usage and minimize the stopwatch's impact on latency.

eguguchkin:
Fixes #158 

---

### DIFF
--- a/frac/processor/search.go
+++ b/frac/processor/search.go
@@ -147,9 +147,13 @@ func iterateEvalTree(
 	ids := seq.IDSources{}
 	var lastID seq.ID
 
-	for {
+	timerEval := sw.Timer("eval_tree_next")
+	timerMID := sw.Timer("get_mid")
+	timerRID := sw.Timer("get_rid")
+	timerAgg := sw.Timer("agg_node_count")
 
-		if util.IsCancelled(ctx) {
+	for i := 0; ; i++ {
+		if i%1000 == 0 && util.IsCancelled(ctx) {
 			return total, ids, histogram, ctx.Err()
 		}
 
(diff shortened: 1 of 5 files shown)


frac/processor/search.go
```diff
@@ -147,9 +147,13 @@ func iterateEvalTree(
...
 
-	for {
+	timerEval := sw.Timer("eval_tree_next")
+	timerMID := sw.Timer("get_mid")
+	timerRID := sw.Timer("get_rid")
+	timerAgg := sw.Timer("agg_node_count")
 
-		if util.IsCancelled(ctx) {
+	for i := 0; ; i++ {
+		if i%1000 == 0 && util.IsCancelled(ctx) {
```
cheb0: Usually, people do it like this: `i & 1023 == 0` instead of `i % 1000 == 0` in hot loops. This is because `%` is considered to be slower than `+` or `*`. However, I found the difference is negligible even in microbenchmark.

metric/stopwatch/stopwatch.go
```diff
@@ -100,3 +98,21 @@ func (sw *Stopwatch) ExportValuesAndCounts(mv, mc *prometheus.HistogramVec, opti
 	}
 	sw.Reset()
 }
+
+type Timer struct {
```
cheb0, line 102: I don't think I could come up with a better name. Maybe, `ElapsedTimer` but I'm not sure.

eguguchkin: There are my variants:

- Meter
- TimeMeter
- Span

eguguchkin: But Timer is also good for me

frac/processor/search.go
eguguchkin: fixed

PR #169, opened 2025-09-30
feat(fracmanager): parallelize seals on replay

### LINKED ISSUE #161: Parallelize fractions seals on replay
Currently, FracManager does a sequential replay and seal of all active fractions after process start. It would be beneficial to parallelize the process for better store startup time.

cheb0:
# Description

Makes frac manager replay and seal fractions in parallel. This will speed up the process of store startup.

Fixes #161

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```


config/config.go
```diff
@@ -169,6 +169,9 @@ type Config struct {
 		// SearchWorkers specifies number of workers for searchers pool.
 		// By default this setting is equal to [runtime.GOMAXPROCS].
 		SearchWorkers int `config:"search_workers"`
+		// ReplayWorkers specifies number of workers.
+		// By default this setting is equal to 8.
+		ReplayWorkers int `config:"replay_workers" default:"8"`
```
dkharms: Why eight workers but not the `GOMAXPROCS` (which usually will be equal to cpu limit in cgrpoups)?

fracmanager/config.go
```diff
@@ -61,6 +62,9 @@ func FillConfigWithDefault(config *Config) *Config {
 	if config.SealParams.TokenTableZstdLevel == 0 {
 		config.SealParams.TokenTableZstdLevel = zstdDefaultLevel
 	}
+	if config.ReplayWorkers == 0 {
+		config.ReplayWorkers = consts.DefaultReplayWorkers
+	}
```
dkharms, line 69: This is not necessary, since you've already set `default` tag in config structure.

Or this is just for tests? If yes then we should explicitly specify the value in test/suite setup routine. We are moving towards removing all of these `if` statements in favor of `default` tag.

fracmanager/fracmanager_test.go
```diff
@@ -158,68 +441,3 @@ func TestNewULID(t *testing.T) {
 	assert.Equal(t, 26, len(ulid1), "ULID should have length 26")
 	assert.Greater(t, ulid2, ulid1)
 }
-
-func TestOldestCT(t *testing.T) {
```
dkharms, line 154: Actually, this test is important and shouldn't be deleted.

fracmanager/fracmanager_test.go
```diff
@@ -93,6 +93,289 @@ func TestCleanUp(t *testing.T) {
 	assert.Equal(t, 1, len(fm.localFracs), "wrong frac count")
 }
 
+func TestReplaySingleEmptyFrac(t *testing.T) {
```
dkharms, line 97: I am curious what is the purpose of these tests? To check that all active fractions except for the youngest one will be sealed? We didn't have tests for that?

config/config.go
eguguchkin: Here we are limited not by cpu-cores but by memory. Therefore, this sensitive parameter must be increased carefully.

eguguchkin: So the default value of 2 seems more appropriate to me

cheb0: Set default value to `2`

fracmanager/config.go
cheb0: There is seq-db config and there is frac manager config. Frac manager can be created from tests, and specifying default value every time seems like too much. So, it seems to me there should be default value here.

fracmanager/fracmanager_test.go
cheb0: I wrote these tests to check that:
- all fractions are sealed
- saved in fm.localFracs in exactly same order. Search will still work if order of fractions doesn't match, but it is important
- tests check that empty fractions are killed
- tests have higher concurrency level with small fractions

In any case, I understand that they may feel redundant, but I it was easier to me to write a new implementation while having them. We can delete some of them after the release

fracmanager/fracmanager_test.go
cheb0: Not on purpose. Restored.

fracmanager/fracmanager.go
```diff
@@ -506,25 +508,75 @@ func (fm *FracManager) Load(ctx context.Context) error {
...
-			return err
+	if len(actives) == 0 {
+		return nil
+	}
+
+	// goroutines access different indices, no need for lock protection for fracRefs
+	var fracRefs = make([]*fracRef, len(actives)-1)
+
+	if len(actives) > 1 {
+		g, ctx := errgroup.WithContext(ctx)
```
dkharms: There is `SetLimit(int)` method on `errgroup` - https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit

> SetLimit limits the number of active goroutines in this group to at most n

It's better to use this API than other package `semaphore`.

fracmanager/fracmanager_test.go
dkharms: Okay, that's fine.

fracmanager/fracmanager.go
```diff
@@ -506,25 +508,75 @@ func (fm *FracManager) Load(ctx context.Context) error {
...
-		if err := a.Replay(ctx); err != nil {
-			return err
+	if len(actives) == 0 {
+		return nil
+	}
+
+	// goroutines access different indices, no need for lock protection for fracRefs
+	var fracRefs = make([]*fracRef, len(actives)-1)
+
+	if len(actives) > 1 {
```
dkharms: What do you think about this?

```go
g, ctx := errgroup.WithContext(ctx)
g.SetLimit(fm.config.ReplayWorkers)

// goroutines access different indices, no need for lock protection for fracRefs
var fracRefs = make([]*fracRef, len(actives)-1)
for i, f := range actives[1:] {
	g.Go(func() error {
		if err := f.Replay(ctx); err != nil {
			return err
		}

		if f.Info().DocsTotal == 0 {
			f.Suicide() // remove empty
			return nil
		}

		ref := fm.newActiveRef(f)
		fracRefs[i] = ref.ref

		fm.seal(ref)
		return nil
	})
}

if err := g.Wait(); err != nil {
	return err
}
```

fracmanager/fracmanager.go
```diff
@@ -506,25 +508,75 @@ func (fm *FracManager) Load(ctx context.Context) error {
...
+	}
 
-		r := fm.newActiveRef(a)
-		fm.localFracs = append(fm.localFracs, r.ref)
+	var active *activeRef
 
-		if i == len(actives)-1 { // last and not empty
-			fm.active = r
-			continue
+	if lastActive.Info().DocsTotal == 0 {
```
dkharms: I do not think that this is necessary, because here you kill the fraction, but later you'll create it anyway (in `fracmanager.Load`):
```go
if fm.active.ref == nil { // no active
	_ = fm.rotate() // make new empty active
}
```

fracmanager/fracmanager.go
cheb0: You are right, I used `setLimit`, works fine

fracmanager/fracmanager.go
cheb0: Looks good. I just replaced `range actives[1:]` with `range actives[:len(actives)-1]`, and it passed all tests.

fracmanager/fracmanager.go
cheb0: That's right. I tried to keep up with the original version, but here we do not need to kill empty active frac. I changed it, and now it looks cleaner.

fracmanager/fracmanager.go
```diff
@@ -506,27 +507,61 @@ func (fm *FracManager) Load(ctx context.Context) error {
...
+
+				fm.seal(ref)
+				return nil
+			})
 		}
 
-		if a.Info().DocsTotal == 0 {
-			a.Suicide() // remove empty
-			continue
+		if err := g.Wait(); err != nil {
```
eguguchkin: nit: To increase parallelism, we could run last.Replay(ctx) concurrently with the other replays. One way to achieve this is by moving g.Wait() after last.Replay(ctx), but this would require decreasing the SetLimit(N) by one.

fracmanager/fracmanager_test.go
```diff
@@ -94,6 +96,305 @@ func TestCleanUp(t *testing.T) {
...
+	dataDir := testscommon.GetTestTmpDir(t)
+	testscommon.RecreateDir(dataDir)
+	defer testscommon.RemoveDir(dataDir)
+
+	fm, err := newFracManagerWithBackgroundStart(t.Context(), &Config{
+		FracSize:     100000000, // maintenance will not seal fracs
+		TotalSize:    100000000,
+		ShouldReplay: false,
```
eguguchkin: It seems we lost track of this option a long time ago. I can't find what it does.

And it's probably better to remove this field altogether rather than start processing it again, when, I assume, it can only be used for testing? (Not in this PR, but it's up to you).

fracmanager/fracmanager_test.go
```diff
@@ -94,6 +96,305 @@ func TestCleanUp(t *testing.T) {
...
+		ShouldReplay:  true,
+		ReplayWorkers: 10,
+		DataDir:       dataDir,
+	})
+	assert.NoError(t, err)
+
+	replayedFracs := fm.getLocalFracs()
+	assert.Equal(t, 1, len(replayedFracs), "should replay exactly one frac")
+	active := fm.Active() // replayed frac is active
+	assert.Equal(t, fractionInfo.DocsOnDisk, active.Info().DocsOnDisk, "should have same doc count for replayed frac")
```
eguguchkin: nit: `DocsOnDisk` represents the size of the docs file in bytes. You should use `DocsTotal` instead. However, the logic of your check is correct; only the message likely needs to be updated.

fracmanager/fracmanager_test.go
```diff
@@ -94,6 +96,305 @@ func TestCleanUp(t *testing.T) {
...
+		addDocs(t, fm, 500+rand.Intn(100))
+		fm.rotate()
+	}
+	addDocs(t, fm, 5)
+
+	var fracs []common.Info
+	for _, fraction := range fm.getLocalFracs() {
+		info := *fraction.Info()
+		if info.DocsTotal == 0 {
+			fmt.Printf("no docs")
```
eguguchkin: This looks like debug code - should it be removed?

fracmanager/fracmanager_test.go
```diff
@@ -94,6 +96,305 @@ func TestCleanUp(t *testing.T) {
...
+		DataDir:       dataDir,
+	})
+	assert.NoError(t, err)
+
+	replayedFracs := fm.getLocalFracs()
+
+	assert.Equal(t, len(fracs), len(replayedFracs), "should replay same number of fractions")
+
+	// all fracs should match exactly (no empty) in same order
+	for i := 0; i < 51; i++ {
```
eguguchkin: fracCount + 1?

fracmanager/fracmanager_test.go
```diff
@@ -94,6 +96,305 @@ func TestCleanUp(t *testing.T) {
...
+		assert.Equal(t, fracs[i].Name(), replayedFracs[i].Info().Name(), "fraction %d should have same name", i)
+		assert.Equal(t, fracs[i].DocsTotal, replayedFracs[i].Info().DocsTotal, "fraction %d should have same doc count", i)
+
+		if i != 50 {
```
eguguchkin: fracCount ?

fracmanager/fracmanager_test.go
```diff
@@ -94,6 +96,305 @@ func TestCleanUp(t *testing.T) {
...
+			assert.Greater(t, replayedFracs[i].Info().SealingTime, uint64(0), "replayed frac %d must be sealed", i)
+		} else {
+			assert.Equal(t, replayedFracs[i].Info().SealingTime, uint64(0), "replayed frac %d must not be sealed", i)
+		}
+	}
+
+	newActive := fm.Active()
+	assert.Greater(t, newActive.Info().DocsTotal, uint32(0), "new active fraction should not be empty")
```
eguguchkin: nit: maybe it would be more accurate to compare with 5? `assert.Equal(t, newActive.Info().DocsTotal, uint32(5)`

fracmanager/fracmanager_test.go
```diff
@@ -94,6 +96,305 @@ func TestCleanUp(t *testing.T) {
...
+		TotalSize:     100000000,
+		ShouldReplay:  false,
+		ReplayWorkers: 10,
+		DataDir:       dataDir,
+	})
+	assert.NoError(t, err)
+
+	for i := 0; i < 50; i++ {
+		addDocs(t, fm, 500+rand.Intn(100))
+		fm.WaitIdle()
```
eguguchkin: Is this necessary? You don't use WaitIdle in other tests. And it seems unnecessary here too.

fracmanager/fracmanager.go
cheb0: Decreasing concurrency by `1` means we also need to change default value for `ReplayWorkers`. If we want to replay an active frac concurrently, I would just move replaying active frac inside `errgroup` and handled it with `if` statement. Think we should do that?

fracmanager/fracmanager_test.go
cheb0: You are right, this field is never read. We can remove it altogether in a different PR (I think it's a "good first issue").

fracmanager/fracmanager_test.go
cheb0: Initially, I used `DocsOnDisk`, but we need to move to `DocsTotal`. `DocsOnDisk` can actually change after seal because of a different compression level.

fracmanager/fracmanager_test.go
cheb0: Removed

fracmanager/fracmanager_test.go
cheb0: Changed

fracmanager/fracmanager_test.go
cheb0: Changed

fracmanager/fracmanager_test.go
cheb0: Fixed

fracmanager/fracmanager_test.go
cheb0: It's not necessary in this particular place, because I added it to `addDocs` method. 

It's actually necessary for tests, because they fail once I remove it. This is because test goroutine might not see `info.DocsTotal` which is updated after data is indexed in a different goroutine. 

And we can't call it only once after appending to all fracs. We have to call it after appending to every fraction, because this method `WaitIdle` waits only for current frac to be idle. If we don't call it after appending to every frac, we can catch behaviour where we waited for a proper `DocsTotal` on the last frac, but not for the previous before the last.

eguguchkin: Yes, `WaitIdle` is needed in general. My comment is specific to this particular place. Especially since it duplicates the call to `addDocs`.

fracmanager/fracmanager.go
cheb0: Implemented parallelize across all fractions.

eguguchkin: Yes, your decision is right!

fracmanager/fracmanager.go
```diff
@@ -506,27 +507,56 @@ func (fm *FracManager) Load(ctx context.Context) error {
...
-		fm.seal(r)
+			return nil
+		})
 	}
 
+	if err := g.Wait(); err != nil {
+		return err
+	}
+
+	var localFracs []*fracRef
```
eguguchkin: nit: Do we need an intermediate slice? We can add it directly to `fm.localFracs`

cheb0: Fixed

fracmanager/fracmanager_test.go
```diff
@@ -94,17 +94,300 @@ func TestCleanUp(t *testing.T) {
...
+	fm, err = newFracManagerWithBackgroundStart(ctx, &Config{
+		FracSize:      100000000,
+		TotalSize:     100000000,
+		ReplayWorkers: 10,
+		DataDir:       dataDir,
+	})
+
+	assert.Error(t, err)
+	assert.Equal(t, err, context.DeadlineExceeded)
+	assert.Equal(t, context.DeadlineExceeded, ctx.Err())
```
forshev: nit: `assert.ErrorIs()`?

fracmanager/fracmanager_test.go
```diff
@@ -94,17 +94,300 @@ func TestCleanUp(t *testing.T) {
...
+		TotalSize:     100000000,
+		ReplayWorkers: 10,
+		DataDir:       dataDir,
+	})
+
+	assert.NoError(t, err)
+
+	replayedFracs := fm.getLocalFracs()
+
+	assert.Equal(t, 5, len(replayedFracs), "only non-empty fracs should remain")
```
forshev: nit: just to clarify things

len of replayedFracs is 5 here since last frac has no docs and `fm.active` is not assigned in the `replayAll()` method because execution for this frac never gets to `if i != len(actives)-1` comparison?

fracmanager/fracmanager_test.go
cheb0: Fixed

fracmanager/fracmanager_test.go
cheb0: Exactly, there are 4 non-empty fracs. The last active frac (empty) is killed in `replayAll` method and a new one is created in fracmanager [Load](https://github.com/ozontech/seq-db/blob/main/fracmanager/fracmanager.go#L489) method.

I think this assert is a little bit ambiguous, so I changed it to 
```
assert.Equal(t, len(nonEmptyFracs)+1, len(replayedFracs), "only non-empty fracs and one active empty frac should remain")
```
and added an additional one which checks that an active frac is empty
```
assert.Equal(t, uint32(0), fm.Active().Info().DocsTotal, "new active fraction should be empty")
```

fracmanager/fracmanager_test.go
```diff
@@ -94,17 +94,301 @@ func TestCleanUp(t *testing.T) {
...
+
+	fm, err = newFracManagerWithBackgroundStart(ctx, &Config{
+		FracSize:      100000000,
+		TotalSize:     100000000,
+		ReplayWorkers: 10,
+		DataDir:       dataDir,
+	})
+
+	assert.Error(t, err)
+	assert.Equal(t, err, context.DeadlineExceeded)
```
forshev: we can use `assert.ErrorIs()` here too

cheb0: I also removed previous assert `assert.Error(t, err)`, since it looks unnecessary with two assert checks that comes later

PR #172, opened 2025-10-02
chore(fraction): get rid of data provider in interface

### LINKED ISSUE #154: Remove frac.DataProvider Interface and Move Methods to fraction.Fraction
Remove the frac.DataProvider interface as it has been simplified to just two methods (Search/Fetch) and is no longer necessary. Move Search/Fetch to fraction.Fraction

eguguchkin:
Fixes #154 

---

### DIFF
--- a/cmd/distribution/main.go
+++ b/cmd/distribution/main.go
@@ -164,7 +164,7 @@ func main() {
 		}
 	}
 
-	fc := fracmanager.NewSealedFracCache(filePathDist)
+	fc := fracmanager.NewFracInfoCache(filePathDist)
 
 	lastSavedTime := time.Now()
 	for _, path := range getAllFracs(dataDir) {
@@ -173,7 +173,7 @@ func main() {
 
 		logger.Info("start process", zap.String("name", key))
 
-		info, ok := fc.GetFracInfo(key)
+		info, ok := fc.Get(key)
 		if ok {
 			logger.Info("found in frac-cache", zap.String("key", key))
 		} else {
@@ -201,7 +201,7 @@ func main() {
 		}
 
 		buildDist(info.Distribution, path, info)
-		fc.AddFraction(key, info)
+		fc.Add(info)
 		logger.Info("built distribution", zap.Int("affected_minutes", len(info.Distribution.GetDist())))
 		printDistribution(info.Distribution)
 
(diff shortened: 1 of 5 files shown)


forshev (approved):

frac/empty_fraction.go
```diff
@@ -0,0 +1,40 @@
...
+}
+func (EmptyFraction) IsIntersecting(seq.MID, seq.MID) bool {
+	return false
+}
+func (EmptyFraction) Contains(mid seq.MID) bool {
+	return false
+}
+
+func (EmptyFraction) Offload(ctx context.Context, u storage.Uploader) (bool, error) {
+	return true, nil
```
dkharms: nit: I guess it's better to return `false` here (since empty fraction cannot be offloaded). But it's up to you since `Offload` method will be removed from `Fraction` interface.

frac/empty.go
```diff
@@ -0,0 +1,40 @@
...
+func (EmptyFraction) Fetch(context.Context, []seq.ID) ([][]byte, error) { return nil, nil }
+
+func (EmptyFraction) Search(_ context.Context, params processor.SearchParams) (*seq.QPR, error) {
+	metric.CountersTotal.WithLabelValues("empty_data_provider").Inc()
+	return &seq.QPR{Aggs: make([]seq.AggregatableSamples, len(params.AggQ))}, nil
+}
+
+func (EmptyFraction) Info() *common.Info {
+	return nil
+}
```
dkharms, line 38: I am wondering - is it safe to return `nil` here? In a lot of places we do not perform `nil` check on `Info` result.

It would be fine if we instantiated `EmptyFraction` only when calling `Search` and `Fetch` (like `EmptyFraction{}.Search(...)`), but there is `cur` method in `proxyFrac` which may return `EmptyFraction` (`proxyFrac.cur().Info().Name()` now might panic).

frac/empty_fraction.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/ozontech/seq-db/storage"
+)
+
+var (
+	_ Fraction = (*EmptyFraction)(nil)
+)
+
+type EmptyFraction struct{}
+
+func (EmptyFraction) Fetch(context.Context, []seq.ID) ([][]byte, error) { return nil, nil }
```
dkharms: What do you think about adding prometheus counter here as well?

```go
metric.CountersTotal.WithLabelValues("empty_data_provider").Inc()
```

PR #174, opened 2025-10-02
fix(async search): fix search request aggs formatting

forshev:
# Description

fixed search request's aggs field formatting for FetchAsycnSearchResult and GetAsyncSearchesList methods

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/proxyapi/grpc_async_search.go
+++ b/proxyapi/grpc_async_search.go
@@ -164,12 +164,20 @@ func (g *grpcV1) DeleteAsyncSearch(
 func makeProtoRequestAggregations(sourceAggs []search.AggQuery) []*seqproxyapi.AggQuery {
 	aggs := make([]*seqproxyapi.AggQuery, 0, len(sourceAggs))
 	for _, agg := range sourceAggs {
-		aggs = append(aggs, &seqproxyapi.AggQuery{
+		agg := &seqproxyapi.AggQuery{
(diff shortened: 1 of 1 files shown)


proxyapi/grpc_async_search.go
```diff
@@ -26,6 +26,11 @@ func (g *grpcV1) StartAsyncSearch(
 			r.Size, g.config.AsyncSearchMaxDocumentsPerRequest)
 	}
 
+	// reject "empty" request: no docs, no aggs, no hist
+	if r.WithDocs == false && len(r.Aggs) == 0 && r.Hist == nil {
+		return nil, status.Error(codes.InvalidArgument, "can't serve empty request: fill aggs, hist or withDocs")
+	}
```
dkharms, line 32: For example, I want to search for documents (I am interested only in `seq.ID`) this will return an error to my request. Is is intended behaviour?

I am asking this because `WithDocs` is just a `bool`, not `*bool`, so it's impossible to distinguish between situations when user specified `false` explicitly and did not provide this option at all.

forshev: yes, this is intended, we discussed it yesterday, we don't want to waste resources on requests that will yield nothing

if you want to save doc ids, you have to set `withDocs=true` explicitly

dkharms (approved):

PR #175, opened 2025-10-02
chore(fracmanager): extract CacheMaintainerMetrics to separate file

### LINKED ISSUE #150: Extract CacheMaintainerMetrics to a Separate File and Add Default Constructor
- Extract CacheMaintainerMetrics into a separate file. 
- Сreate a constructor for default CacheMaintainerMetrics

eguguchkin:
Fixes #150 

---


fracmanager/metrics.go
```diff
@@ -0,0 +1,90 @@
+package fracmanager
```
dkharms, line 1: So let's clarify this. We decided to follow this approach? Place metrics as near as possible to the place where they are used but in the separate file with name `metrics.go`?

moflotas (approved):

fracmanager/metrics.go
eguguchkin: We split into those who wanted to declare metrics locally and those who wanted to put them in a single file. But there were no strong advocates for either approach. Everyone was willing to compromise. We chose a balanced solution - to have a `metrics.go` in each package.

PR #178, opened 2025-10-05
fix: use default docblock size > 0

eguguchkin:


eguguchkin:
> There is also a new parameter which was introduced lately: DocBlocksZstdLevel
> 
> Currently, in tests it is not initialized, and 0 is used as a compression level. Shouldn't it be initialized to default somewhere here? https://github.com/ozontech/seq-db/blob/main/fracmanager/config.go#L62
> 
> UPD: DocBlockSize might also need to be initialized there

I used the same approach as in the previous version of the sealing. `FillConfigWithDefault` is a good practice, but in our current implementation, it does not guarantee that the configuration passed to the sealing has been validated.

So I just solved the local problem in place

moflotas (approved):

PR #181, opened 2025-10-07
chore: make existing token extraction code reusable

eguguchkin:
Fixes #180 

---

- [x] I have read and followed all requirements in [CONTRIBUTING.md](https://github.com/ozontech/seq-db/blob/main/.github/CONTRIBUTING.md);
- [ ] I used LLM/AI assistance to make this pull request;

PR #81, opened 2025-08-15
chore(parser): remove old parser

moflotas:
Fixes # (76)

PR #185, opened 2025-10-10
chore: fraction search unit tests for active/sealed fraction

### LINKED ISSUE #177: Add unit tests for fraction search
Currently, most fraction search tests are integration tests which reside in `tests/integration_tests/single_test.go`. We need to add unit tests for fraction (active, sealed) which do not start dedicated proxy and store instances.

Scenario for tests:
- OR / AND / NOT queries
- wildcards
- limit / offset / (with and without total)
- asc/desc sort
- hists
- aggregations

We need to run tests against different suites:
- active fraction
- freshly sealed fraction
- sealed and read from disk fraction

cheb0:
# Description

This pull requests migrates a part of integration tests for fraction search to unit tests.

Fixes #177

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/docs/en/01-quickstart.md
+++ b/docs/en/01-quickstart.md
@@ -22,6 +22,9 @@ Before launch you need to create a config file:
 config.yaml:
 
 ```yaml
+storage:
+  data_dir: /seq-db-data
+
 mapping:
   path: auto
 ```
@@ -33,6 +36,7 @@ docker run --rm \
   -p 9002:9002 \
   -p 9004:9004 \
   -p 9200:9200 \
+  -v "$(pwd)"/seq-db-data:/seq-db-data \
   -v "$(pwd)"/config.yaml:/seq-db/config.yaml \
   -it ghcr.io/ozontech/seq-db:latest --mode single --config=config.yaml
 ```
--- a/docs/ru/01-quickstart.md
+++ b/docs/ru/01-quickstart.md
@@ -22,6 +22,9 @@ seq-db можно быстро запустить в docker-контейнере
 config.yaml:
 
 ```yaml
+storage:
+  data_dir: /seq-db-data
+
 mapping:
   path: auto
 ```
@@ -33,6 +36,7 @@ docker run --rm \
   -p 9002:9002 \
   -p 9004:9004 \
   -p 9200:9200 \
+  -v "$(pwd)"/seq-db-data:/seq-db-data \
   -v "$(pwd)"/config.yaml:/seq-db/config.yaml \
   -it ghcr.io/ozontech/seq-db:latest --mode single --config=config.yaml
 ```


frac/fraction_test.go
```diff
@@ -0,0 +1,1467 @@
...
+	"github.com/stretchr/testify/suite"
+)
+
+type FractionTestSuite struct {
+	suite.Suite
+	tmpDir         string
+	config         *Config
+	mapping        seq.Mapping
+	tokenizers     map[seq.TokenizerType]tokenizer.Tokenizer
+	activeIndexers []*ActiveIndexer
```
eguguchkin: You can share one indexer for all factions

frac/fraction_test.go
```diff
@@ -0,0 +1,1467 @@
...
+		DocBlockSize:           128 * int(units.KiB),
+	}
+
+	var err error
+	s.tmpDir, err = os.MkdirTemp("", "fraction_test_*")
+	s.Require().NoError(err)
+}
+
+func newSmallCache[V any]() *cache.Cache[V] {
+	return cache.NewCache[V](cache.NewCleaner(uint64(units.KiB), nil), nil)
```
eguguchkin: In this case, the cleaner isn't used for clearing, so it can be replaced with nil – it will work just as well and even simpler.

Do we want to implement cache clearing here? I wouldn't do that at this stage.

frac/fraction_test.go
```diff
@@ -0,0 +1,1467 @@
...
+				slices.Reverse(docs)
+			}
+
+			fetchedDocs := make([]string, 0, len(docs))
+			for _, doc := range docs {
+				fetchedDocs = append(fetchedDocs, string(doc))
+			}
+
+			for i, fetchedDoc := range fetchedDocs {
+				if i < len(expectedIndexes) {
```
eguguchkin: If we remove this check, will we miss the case where we find more documents than expected?

But this shouldn't happen anyway, since we're checking above.

`s.Require().Equal(len(expectedIndexes), qpr.IDs.Len(), "doc count does not match")`

frac/fraction_test.go
```diff
@@ -0,0 +1,1467 @@
...
+	params *processor.SearchParams,
+	originalDocs []string,
+	expectedIndexes []int,
+	checkTotal bool) {
+
+	var withTotals = []bool{false}
+
+	// We can check total only if limit is not set. Otherwise, total returns a count
+	// of all docs which match the query
+	if checkTotal && params.Limit == math.MaxInt32 {
```
eguguchkin: Let's remove the Total check mode from AssertSearchWithSearchParams

And we'll make the Total test itself a separate, simple test.

frac/fraction_test.go
```diff
@@ -0,0 +1,1467 @@
...
+		s.Require().NoError(err, "processing bulk failed")
+
+		compressor.CompressDocsAndMetas(binaryDocs, binaryMeta)
+		docsBlock, metasBlock := compressor.DocsMetas()
+
+		var wg sync.WaitGroup
+		wg.Add(1)
+		err = active.Append(docsBlock, metasBlock, &wg)
+		s.Require().NoError(err, "append to active failed")
+		wg.Wait()
```
eguguchkin: It's better to move Wait out of the loop, then we'll get the advantage of asynchronous indexing (Append will work asynchronously)

frac/fraction_test.go
```diff
@@ -0,0 +1,1467 @@
...
+func (s *SealedLoadedFractionTestSuite) TearDownTest() {
+	if s.fraction != nil {
+		s.fraction.Suicide()
+		s.fraction = nil
+	}
+	s.TearDownTestCommon()
+}
+
+func TestFractionSuites(t *testing.T) {
+	suite.Run(t, new(ActiveFractionTestSuite))
```
eguguchkin, line 1483: nit: Overall, this approach works, but it comes at a cost: re-indexing all the bulks being tested each time. Four times for each test case: for the active one, for the sealed one, for the fraction from disk, and for the remote one.

I like the approach from single_test.go, where all test cases are defined in a function that is run through the RunFracEnvs wrapper.

But I'd just pass a function with a parameter `Fraction`, something like

```go
testCase := func(f Fraction) {
// test cases
}
```

Then we index once and run the test case four times (changing the fraction state).

frac/fraction_test.go
```diff
@@ -0,0 +1,1467 @@
...
+		s.Require().True(info.IndexOnDisk > uint64(1450) && info.IndexOnDisk < uint64(1500),
+			"index on disk doesn't match. actual value: %d", info.MetaOnDisk)
+	default:
+		s.Require().Fail("unsupported fraction type")
+	}
+}
+
+type searchOption func(*processor.SearchParams) error
+
+func (s *FractionTestSuite) query(queryString string, options ...searchOption) *processor.SearchParams {
```
eguguchkin, line 1064: Here (https://github.com/ozontech/seq-db/pull/81), we plan to get rid of `parser.ParseQuery` so that only `parser.ParseSeqQ` remains relevant.

frac/fraction_test.go
```diff
@@ -0,0 +1,1467 @@
...
+		`{"timestamp":"2000-01-01T13:00:05Z","service":"service_c","message":"cache hit","source":"prod03","level":"3"}`,
+		`{"timestamp":"2000-01-01T13:00:06Z","service":"service_c","message":"processing request","source":"prod01","level":"2"}`,
+		`{"timestamp":"2000-01-01T13:00:07Z","service":"service_a","message":"request failed","source":"prod02","level":"1"}`,
+		`{"timestamp":"2000-01-01T13:00:08Z","service":"service_b","message":"processing failed","source":"prod03","level":"4"}`,
+		`{"timestamp":"2000-01-01T13:00:09Z","service":"service_b","message":"processing retry","source":"prod03","level":"3"}`,
+	}
+	var bulk1 []string
+	var bulk2 []string
+	var bulk3 []string
+	docs = append(docs, bulk1...)
```
eguguchkin, line 940: looks unnecessary

frac/fraction_test.go
cheb0: Great. It allowed to reduce test runtime by ~50 millis.

frac/fraction_test.go
cheb0: Ok, set cleaner as `nil`. Deleted this method `newSmallCache` as it is not necessary anymore.

frac/fraction_test.go
cheb0: Removed

frac/fraction_test.go
cheb0: yes, removed

frac/fraction_test.go
cheb0: I removed all queries which are not supported in seq-ql. This test only uses seq-ql now.

frac/fraction_test.go
cheb0: We discussed and decided to stay with the current approach, since perf is not a problem right now. This test takes 750-800 millis to fully complete now on my PC, which can be improved to ~500 millis with disabling fsync, and even further down to 250 millis if temp dir is created in `/dev/shm`

frac/fraction_test.go
cheb0: Done

frac/fraction_test.go
cheb0: Removed this total check and added `TestSearchWithTotal`

frac/fraction_test.go
```diff
@@ -0,0 +1,1449 @@
...
+	docs := []string{
+		/*0*/ `{"timestamp":"2000-01-01T13:00:00.000Z","spans":[{"span_id":"1"},{"span_id":"2"}]}`,
+		/*1*/ `{"timestamp":"2000-01-01T13:00:00.001Z","spans":[{"span_id":"2"},{"span_id":"3"}]}`,
+		/*2*/ `{"timestamp":"2000-01-01T13:00:00.002Z","spans":[{"span_id":"1"},{"span_id":"3"}]}`,
+		/*3*/ `{"timestamp":"2000-01-01T13:00:00.003Z","spans":[{"span_id":"4"},{"span_id":"5"}]}`,
+	}
+
+	s.insertDocuments(docs)
+
+	s.AssertSearch("spans.span_id:*", docs, []int{3, 2, 1, 0})
```
cheb0, line 396: Btw, It seems there is a bug here. This query `spans.span_id:*` returns total `8`, while there are only 4 docs exist.

frac/fraction_test.go
```diff
@@ -0,0 +1,1449 @@
+package frac
+
+import (
```
dkharms, line 3: Please run `make imports`.

frac/fraction_test.go
```diff
@@ -0,0 +1,1449 @@
...
+	s.activeIndexer.Stop()
+}
+
+func (s *FractionTestSuite) SetupTestCommon() {
+	s.config = &Config{
+		Search: SearchConfig{
+			AggLimits: AggLimits{
+				MaxFieldTokens:     1000,
+				MaxGroupTokens:     1000,
+				MaxTIDsPerFraction: 1000,
```
dkharms: IMO in tests these settings must be disabled (set to zero).

frac/fraction_test.go
```diff
@@ -0,0 +1,1449 @@
...
+		LIDsZstdLevel:          1,
+		TokenListZstdLevel:     1,
+		DocsPositionsZstdLevel: 1,
+		TokenTableZstdLevel:    1,
+		DocBlocksZstdLevel:     1,
+		DocBlockSize:           128 * int(units.KiB),
+	}
+
+	var err error
+	s.tmpDir, err = os.MkdirTemp("", "fraction_test_*")
```
dkharms: nit: It's always bugging me that `os.MkdirTemp` accepts empty dir and in this case overrides it to `os.TempDir`. So I always write `os.MkdirTemp(os.TempDir(), "fraction_test_*")` to be more explicit.

frac/fraction_test.go
```diff
@@ -0,0 +1,1449 @@
...
+		cache.NewCache[[]byte](nil, nil),
+		cache.NewCache[[]byte](nil, nil),
+		s.config,
+	)
+
+	var wg sync.WaitGroup
+
+	for _, docs := range bulks {
+		docsCopy := make([]string, len(docs))
+		copy(docsCopy, docs)
```
dkharms: nit: `slices.Clone`

frac/fraction_test.go
```diff
@@ -0,0 +1,1449 @@
...
+			return d, nil
+		}
+
+		proc := indexer.NewProcessor(s.mapping, s.tokenizers, 0, 0, 0)
+		compressor := indexer.GetDocsMetasCompressor(3, 3)
+		_, binaryDocs, binaryMeta, err := proc.ProcessBulk(time.Now(), nil, nil, readNext)
+		s.Require().NoError(err, "processing bulk failed")
+
+		compressor.CompressDocsAndMetas(binaryDocs, binaryMeta)
+		docsBlock, metasBlock := compressor.DocsMetas()
```
dkharms, line 1279: I am not sure that I am fine with this. It's basically a copy paste from [`proxy/bulk/ingestor.go`](https://github.com/ozontech/seq-db/blob/cf66789ac9257d3d8ce8b031eb13547a14cda823/proxy/bulk/ingestor.go#L168) and I am not sure that _in this case_ it is a good idea.

frac/fraction_test.go
```diff
@@ -0,0 +1,1449 @@
...
+
+func TestActiveFractionTestSuite(t *testing.T) {
+	suite.Run(t, new(ActiveFractionTestSuite))
+}
+
+func TestSealedFractionTestSuite(t *testing.T) {
+	suite.Run(t, new(SealedFractionTestSuite))
+}
+
+func TestSealedLoadedFractionTestSuite(t *testing.T) {
```
dkharms, line 1494: Since you've added test for this case does it make sense to add similar test case for an active fraction? When we load data from disk and replay `.meta` file to memory.

frac/fraction_test.go
```diff
@@ -0,0 +1,1449 @@
...
+	s.AssertSearch(s.query("level:5", withLimit(100)), docs, level5Indexes[:100])
+}
+
+func (s *FractionTestSuite) TestFractionInfo() {
+	docs := []string{
+		`{"timestamp":"2000-01-01T13:00:25Z","service":"service_a","message":"first message some text", "service":"gateway"}`,
+		`{"timestamp":"2000-01-01T13:00:32Z","service":"service_b","message":"second message other text", "service":"kube-proxy"}`,
+		`{"timestamp":"2000-01-01T13:00:43Z","service":"service_c","message":"third message other text", "service":"gateway"}`,
+		`{"timestamp":"2000-01-01T13:00:53Z","service":"service_a","message":"fourth message some text", "service":"kube-proxy"}`,
+		`{"timestamp":"2000-01-01T13:00:54Z","service":"service_c","message":"apple","service":"kube-scheduler"}`,
```
dkharms: Here `service` field is duplicated. Has that made intentionally?

frac/fraction_test.go
eguguchkin: This can be fixed, but at the cost of additional disk reads. We've put off solving this problem for now. But I'll create an issue.

eguguchkin: #215

frac/fraction_test.go
eguguchkin: What's wrong with this approach?

frac/fraction_test.go
cheb0: Fixed

frac/fraction_test.go
cheb0: Fixed

frac/fraction_test.go
cheb0: Agreed, `os.MkdirTemp("", "fraction_test_*")` does not look so obvious

frac/fraction_test.go
cheb0: Fixed

frac/fraction_test.go
cheb0: Currently, this piece of code doesn't look problematic. But it will become worse if it's copied across the project. As far as, one option is to augment `TestDocProvider` so that it supports mapping and tokenizers, which will reduce amount of boilerplate.

IMO, it's not about copy-paste/beauty/whatsoever. It's about having an API which allows you to create an active frac for test in 2-3 lines of code. If it's possible, writing unit tests will become easier.

frac/fraction_test.go
cheb0: Added a new suite `ActiveReplayedFractionTestSuite`

frac/fraction_test.go
cheb0: Nope, fixed

frac/fraction_test.go
eguguchkin: Historically, we haven't had such an API. Some issues were resolved in https://github.com/ozontech/seq-db/pull/181, but no new API was created in that PR. I think it makes sense to address this issue in a separate task.

#216

PR #186, opened 2025-10-11
fix: sealer memory leak into preloaded data

eguguchkin:
Holding a pointer to a part of the short-lived IndexSealer structure keeps all temporary buffers of this structure alive.

---

### DIFF
--- a/frac/sealed/token/table_loader.go
+++ b/frac/sealed/token/table_loader.go
@@ -121,11 +121,21 @@ func (b TableBlock) packedSize() int {
 	const sizeOfUint32 = int(unsafe.Sizeof(uint32(0)))
 	size := 0
 	for _, fieldData := range b.FieldsTables {
-		size += 2 * sizeOfUint32
+		// field name
+		size += sizeOfUint32
 		size += len(fieldData.Field)
+		// entries count
+		size += sizeOfUint32
 		for _, entry := range fieldData.Entries {
-			size += 8 * sizeOfUint32
+			size += sizeOfUint32
+			size += sizeOfUint32
+			size += sizeOfUint32
+			size += sizeOfUint32
+			// MinVal
+			size += sizeOfUint32
 			size += len(entry.MinVal)
+			// MaxVal
+			size += sizeOfUint32
 			size += len(entry.MaxVal)
 		}
 	}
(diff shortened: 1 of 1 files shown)


frac/sealed/sealing/index.go
```diff
@@ -432,8 +434,8 @@ func (s *IndexSealer) packLIDsBlock(block lidsSealBlock) indexBlock {
 }
 
 // LIDsTable returns the built LIDs table for fast initialization of sealed fraction.
-func (s *IndexSealer) LIDsTable() *lids.Table {
-	return &s.lidsTable
+func (s *IndexSealer) LIDsTable() lids.Table {
```
eguguchkin, line 437: Here it is!

dkharms: Let me expand your explanation (and please correct if I'm wrong in some details).

So for each invocation of `sealing.Seal` we allocate short-lived `IndexSealer` struct (which holds `lidsTable`). But `lidsTable` outlives `IndexSealer` (because we keep preloaded data for sealed fractions in memory and `lidsTable` is a part of that preloaded data).

So when we return `&s.lidsTable` what happens is that we return a pointer to the slice which is stored *inside* `IndexSealer` instance. And since pointer is reachable while sealed fraction is alive (not suicided) GC cannot collect instance of `IndexSealer` which also has two temporary buffers (each takes 16KiB).

frac/sealed/token/table_loader.go
```diff
@@ -115,7 +117,24 @@ type FieldTable struct {
...
 }
 
+func (b TableBlock) packedSize() int {
+	const sizeOfUint32 = int(unsafe.Sizeof(uint32(0)))
+	size := 0
+	for _, fieldData := range b.FieldsTables {
+		size += 2 * sizeOfUint32
+		size += len(fieldData.Field)
+		for _, entry := range fieldData.Entries {
+			size += 8 * sizeOfUint32
```
dkharms: I believe here must be `size += 6 * sizeOfUint32` because in the `Pack()` method we do the following:
```go
for _, entry := range fieldData.Entries {
	buf = binary.LittleEndian.AppendUint32(buf, entry.StartTID) // #1
	buf = binary.LittleEndian.AppendUint32(buf, entry.ValCount) // #2
	buf = binary.LittleEndian.AppendUint32(buf, entry.StartIndex) // #3
	buf = binary.LittleEndian.AppendUint32(buf, entry.BlockIndex) // #4
	// MinVal
	buf = binary.LittleEndian.AppendUint32(buf, uint32(len(entry.MinVal))) // #5
	buf = append(buf, entry.MinVal...) // not important here
	// MaxVal
	buf = binary.LittleEndian.AppendUint32(buf, uint32(len(entry.MaxVal))) // #6
	buf = append(buf, entry.MaxVal...) // not important here
}
```

But it's difficult to reason about these constants, maybe you should split addition into several lines to match the code structure in `Pack()` method (I am 99.99% sure that compiler will optimize several additions into one addition so there will be no performance penalty)? Or at least add comments.

frac/sealed/sealing/index.go
eguguchkin: Yes, that's right!

frac/sealed/token/table_loader.go
eguguchkin: fixed it

PR #193, opened 2025-10-13
refactor(fracmanager): move replaying and sealing to the loader

### LINKED ISSUE #152: Move Replay and Seal to Loader and Rename fracInfo to fracManifest
Move Replay and Seal to Loader. Rename the type fracInfo to fracManifest (to avoid collision with common.Info) and extract it into a separate file

eguguchkin:
Fixes #152 

---

### DIFF
--- a/fracmanager/frac_manifest.go
+++ b/fracmanager/frac_manifest.go
@@ -72,11 +72,11 @@ func (m *fracManifest) AddExtension(ext string) error {
 type fracStage int
 
 const (
-	fracStageActive  fracStage = iota // active fraction - accepts new data
+	fracStageUnknown fracStage = iota // unknown state - requires analysis
+	fracStageActive                   // active fraction - accepts new data
 	fracStageSealed                   // sealed fraction - read-only
 	fracStageRemote                   // remote fraction - data stored in external storage
 	fracStageZombie                   // partially deleted fraction - requires cleanup
-	fracStageUnknown                  // unknown state - requires analysis
 )
 
 // Stage determines the current stage of the fraction based on file presence
@@ -184,9 +184,7 @@ func analyzeFiles(files []string) ([]*fracManifest, error) {
 		}
 	}
 
-	sort.Strings(ids) // sort by identifiers
-
-	// Filter valid fractions
+	sort.Strings(ids)
 	return filterValid(ids, manifests)
 }
 
@@ -244,7 +242,7 @@ func cleanupRemoteFrac(m *fracManifest) {
 func cleanupSealedFrac(m *fracManifest) {
 	removeMeta(m)
 	if m.hasSdocs {
-		removeDocs(m) // remove orig docs, keep sorted
+		removeDocs(m) // remove orig docs, but keeping sorted
 	}
 }
 
--- a/fracmanager/loader.go
+++ b/fracmanager/loader.go
@@ -21,8 +21,8 @@ type Loader struct {
 	infoCache *fracInfoCache    // fraction metadata cache
 
 	cacheStat struct {
-		hits   int // counter of fractions loaded from cache
-		misses int // counter of fractions loaded without using cache
+		hits   int // counter of fractions loaded from frac info cache
+		misses int // counter of fractions loaded without using frac info cache
 	}
 }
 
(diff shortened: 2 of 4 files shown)


fracmanager/frac_manifest.go
```diff
@@ -0,0 +1,312 @@
...
+	hasIndexDel bool // index deletion marker
+
+	// Temporary file flags
+	hasIndexTmp bool // temporary index file
+	hasSdocsTmp bool // temporary sorted documents file
+}
+
+// AddExtension adds information about a file with the specified extension
+// Called during filesystem analysis to build a complete picture of fraction files
+func (m *fracManifest) AddExtension(ext string) error {
```
cheb0, line 39: nit: could be `addExtension`

fracmanager/frac_manifest.go
```diff
@@ -0,0 +1,312 @@
...
+	fracStageActive  fracStage = iota // active fraction - accepts new data
+	fracStageSealed                   // sealed fraction - read-only
+	fracStageRemote                   // remote fraction - data stored in external storage
+	fracStageZombie                   // partially deleted fraction - requires cleanup
+	fracStageUnknown                  // unknown state - requires analysis
+)
+
+// Stage determines the current stage of the fraction based on file presence
+// Key method for making fraction management decisions
+func (m *fracManifest) Stage() fracStage {
```
cheb0, line 84: nit: `stage`?

fracmanager/loader.go
```diff
@@ -3,273 +3,222 @@ package fracmanager
...
-		if c == '.' {
-			return str[i:]
-		}
+// logLoadingProgress logs loading progress at regular intervals
+// Provides visibility into the fraction loading process
+func logLoadingProgress(startTime time.Time, currentIndex, totalCount int) {
+	if time.Since(startTime) >= time.Second || currentIndex == totalCount-1 {
+		progressPercent := 100 * (currentIndex + 1) / totalCount
+		logger.Info(
+			"fraction loading progress",
```
cheb0: nit: maybe `fraction list loading progress`?

Right now, if I have 10 active fractions on disk, this message is printed almost instantly with 100% progress result after a store had started. However, replay then can take seconds even on a local setup. What I am saying, we have a progress reporting here which can be not so obvious for everybody else. With a different message it's easier to grasp the meaning.

fracmanager/loader.go
```diff
@@ -3,273 +3,222 @@ package fracmanager
...
+// Coordinates the process of discovering, validating, and loading all fraction types
+type Loader struct {
+	config    *Config           // loader configuration
+	provider  *fractionProvider // provider for creating fraction objects
+	infoCache *fracInfoCache    // fraction metadata cache
 
-	cachedFracs   int
-	uncachedFracs int
+	cacheStat struct {
+		hits   int // counter of fractions loaded from cache
```
cheb0: nit: I think the comment can be `...from frac info cache`, but struct name can stay the same. that's because we already have the other (in-memory) `cache` which is made for a different purpose.

fracmanager/loader.go
```diff
@@ -3,273 +3,222 @@ package fracmanager
...
 
-func (l *loader) filterInfos(fracIDs []string, infos map[string]*fracInfo) []*fracInfo {
-	infoList := make([]*fracInfo, 0)
-
-	for _, id := range fracIDs {
-		info := infos[id]
-		if info == nil {
-			logger.Panic("frac loader has gone crazy")
+	// Apply load limit if specified in configuration
+	if l.config.FracLoadLimit > 0 {
```
dkharms: This option is always equal to zero. Do we really need it?

fracmanager/frac_manifest.go
```diff
@@ -0,0 +1,312 @@
...
+// Used to determine fraction state and corresponding cleanup operations
+type fracStage int
+
+const (
+	fracStageActive  fracStage = iota // active fraction - accepts new data
+	fracStageSealed                   // sealed fraction - read-only
+	fracStageRemote                   // remote fraction - data stored in external storage
+	fracStageZombie                   // partially deleted fraction - requires cleanup
+	fracStageUnknown                  // unknown state - requires analysis
+)
```
dkharms, line 80: It's better to place unspecified or unknown states to the beginning.

We do not have a good default here and it will help to find places where we could occasionally forget to specify stage (it would unknown instead of active).

dkharms (approved):

fracmanager/frac_manifest.go
eguguchkin: Well, some time ago in our inner documentation we added this:

> Public methods should be used in private structures to indicate which methods encapsulate the logic of that structure.

We don't have it in our .github/CONTRIBUTING.md right now, so I think we need to decide on it again.

fracmanager/frac_manifest.go
eguguchkin: Same as above

fracmanager/frac_manifest.go
eguguchkin: done

fracmanager/loader.go
eguguchkin: done

fracmanager/loader.go
eguguchkin: done

fracmanager/loader.go
eguguchkin: done (removed)

PR #196, opened 2025-10-14
refactor(fracmanager): move immature flag logic to store state manager

### LINKED ISSUE #153: Extract Immature Flag Logic into a Dedicated StateManager Type
Extract the logic for managing the immature flag into a separate type called StateManager and rename the flag to CapacityExceeded

eguguchkin:
Fixes #153 

---

### DIFF
--- a/fracmanager/storage_state.go
+++ b/fracmanager/storage_state.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+
+	"github.com/ozontech/seq-db/util"
 )
 
 const StateFile = "storage-state.json"
@@ -16,14 +18,12 @@ type StorageState struct {
 
 // StateManager manages storage state with thread safety
 type StateManager struct {
-	mu       sync.Mutex
+	mu       sync.RWMutex
 	current  StorageState
 	filePath string
 }
 
 // NewStateManager creates a new storage state manager
-// dataDir - directory for storing state file
-// defaultState - default state if file doesn't exist
 func NewStateManager(dataDir string, defaultState StorageState) (*StateManager, error) {
 	sm := &StateManager{
 		filePath: filepath.Join(dataDir, StateFile),
@@ -34,8 +34,8 @@ func NewStateManager(dataDir string, defaultState StorageState) (*StateManager, 
 
 // IsCapacityExceeded returns storage capacity exceeded flag
 func (m *StateManager) IsCapacityExceeded() bool {
-	m.mu.Lock()         // Lock for thread safety
-	defer m.mu.Unlock() // Ensure unlock
+	m.mu.RLock()
+	defer m.mu.RUnlock()
 	return m.current.CapacityExceeded
 }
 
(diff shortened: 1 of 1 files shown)


fracmanager/storage_state.go
```diff
@@ -0,0 +1,142 @@
...
+
+const StateFile = "storage-state.json"
+
+type StorageState struct {
+	CapacityExceeded bool `json:"capacity_exceeded"` // Storage capacity exceeded flag
+}
+
+// StateManager manages storage state with thread safety
+type StateManager struct {
+	mu       sync.Mutex
```
moflotas: nit: I assume, that in future we would use this file for more such flags. In such case, I expect more reads than writes, so it might be preferable to use RWMutex, but generally the contention would not be that huge, and we can ignore that

fracmanager/fracmanager.go
```diff
@@ -239,8 +245,10 @@ func (fm *FracManager) determineOutsiders() []frac.Fraction {
 		truncated++
 	}
 
-	if len(outsiders) > 0 && !fm.Mature() {
-		fm.setMature()
+	if len(outsiders) > 0 && !fm.flags.IsCapacityExceeded() {
+		if err := fm.flags.setCapacityExceeded(true); err != nil {
+			logger.Fatal("set capacity exceeded error", zap.Error(err))
+		}
```
moflotas, line 251: nit: maybe name method `mustSetCapacityExceeded` which will check error inside, to not bloat the code

fracmanager/storage_state.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// Sync parent directory
+	if err = syncDir(filepath.Dir(path)); err != nil {
+		return fmt.Errorf("sync dir: %w", err)
+	}
+
+	return nil
+}
+
+func syncDir(dir string) error {
```
moflotas: There is `MustSyncPath` function in `util/fs.go`, which does the same (or if you want to return an error, `SyncPath` function also exists)

fracmanager/storage_state.go
eguguchkin: Currently, we write this flag once during the entire life of the store and read it twice per minute from a single goroutine.

But expanding this in the future might be a good idea.

fracmanager/fracmanager.go
eguguchkin: I don't want to make the StateManager responsible for stopping execution. Let it return an error.

fracmanager/storage_state.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return fmt.Errorf("create temp file: %w", err)
+	}
+
+	defer func() {
+		if f != nil {
+			f.Close() // Close file if still open
+		}
+		if err != nil {
+			os.Remove(tmpPath) // Remove temp file on error
```
cheb0: nit: I think this comment (and some more in this file) states exactly what the code is doing. More comments doesn't necessarily mean more readable code

fracmanager/storage_state.go
eguguchkin: done

fracmanager/storage_state.go
eguguchkin: done

fracmanager/storage_state.go
eguguchkin: deleted redundant comments

PR #222, opened 2025-10-26
docs: basic config.yaml sample extended with "storage.data_dir" option

Alhanaqtah:
# Description
I tried to run single mode seq-db using the configuration file from the documentation, but I ran into an error: ```config validation failed","error":"field \"storage.data_dir\" is required``` Then I found out that specifying "storage.data_dir" is just as necessary for startup as "mapping.path",  so I decided to add it to complement the example from the documentation.

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/docs/en/01-quickstart.md
+++ b/docs/en/01-quickstart.md
@@ -22,6 +22,9 @@ Before launch you need to create a config file:
 config.yaml:
 
 ```yaml
+storage:
+  data_dir: /seq-db-data
+
 mapping:
   path: auto
 ```
@@ -33,6 +36,7 @@ docker run --rm \
   -p 9002:9002 \
   -p 9004:9004 \
   -p 9200:9200 \
+  -v "$(pwd)"/seq-db-data:/seq-db-data \
   -v "$(pwd)"/config.yaml:/seq-db/config.yaml \
   -it ghcr.io/ozontech/seq-db:latest --mode single --config=config.yaml
 ```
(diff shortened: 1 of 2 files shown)


docs/en/01-quickstart.md
```diff
@@ -22,6 +22,9 @@ Before launch you need to create a config file:
 config.yaml:
 
 ```yaml
+storage:
```
cheb0, line 25: I'd suggest to to use a different folder for data. 

```
storage:
    data_dir: /seq-db-data
```

And then update the `docker run` command with corresponding volume mapping:

```
docker run --rm \
  -p 9002:9002 \
  -p 9004:9004 \
  -p 9200:9200 \
  -v "$(pwd)"/seq-db-data:/seq-db-data \
  -v "$(pwd)"/config.yaml:/seq-db/config.yaml \
  -it ghcr.io/ozontech/seq-db:latest --mode single --config=config.yaml
```

I think it's more cleaner approach. It also allows first-time users to inspect the resulting files after posing docs. Otherwise, potential users may stick to this command for a long time and may use `seq-db` folder as a fraction (shard) storage which itself may have unwanted errors in logs.

Alhanaqtah: Thanks for your review! I think your approach is better. I'll fix that

dkharms:
@Alhanaqtah thanks for your contribution! Merging.

dkharms (approved):

PR #230, opened 2025-10-27
chore(tests): add a concurrency test for active fraction append/search

### LINKED ISSUE #218: Add concurrency tests for active fraction
Add test which will concurrently write and read to active fraction. The test can later be executed with `-race` flag in CI to detect data races.

cheb0:
# Description

Simple concurrency test for an active fraction. 

_Write path_: we create bulks of messages, shuffle them and send them concurrently (8 threads). Each goroutine occasionally issues queries but does not check QPR results.
 
_Read path_: we issue queries concurrently (8 threads). A query is picked in a random fashion, as well as timestamp. Fetched docs are matched exact the original docs.

It takes ~2 seconds with race detector enabled to execute the test.

Fixes #218

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```


frac/active_concurrency_test.go
```diff
@@ -0,0 +1,281 @@
...
+	const numIndexWorkers = 8
+	const numWriters = 8
+	const numReaders = 8
+	const numQueries = 500
+	const numMessagesPerWriter = 5000
+	const bulkSize = 100
+
+	docs, bulks, fromTime, toTime := generatesMessages(numWriters*numMessagesPerWriter, bulkSize)
+
+	tmpDir := common.CreateTempDir()
```
eguguchkin: The common.CreateTempDir() function always returns the same directory path at runtime. Consequently, tests will fail when using the `-count N` flag since they'll attempt to reuse the same directory.

Why not use `baseName := filepath.Join(t.TempDir(), "test_fraction")` instead.

frac/active_concurrency_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+	activeIndexer := NewActiveIndexer(numIndexWorkers, 1000)
+	activeIndexer.Start()
+	defer activeIndexer.Stop()
+
+	fraction := NewActive(
+		baseName,
+		activeIndexer,
+		storage.NewReadLimiter(1, nil),
+		cache.NewCache[[]byte](nil, nil),
```
eguguchkin, line 53: nit: I would use cache cleaner in this test

`cleaner := cache.NewCleaner(4096, nil)`

frac/active_concurrency_test.go
```diff
@@ -0,0 +1,281 @@
...
+	baseName := filepath.Join(tmpDir, "test_fraction")
+
+	activeIndexer := NewActiveIndexer(numIndexWorkers, 1000)
+	activeIndexer.Start()
+	defer activeIndexer.Stop()
+
+	fraction := NewActive(
+		baseName,
+		activeIndexer,
+		storage.NewReadLimiter(1, nil),
```
eguguchkin: For true concurrency i would set `maxReadsNum` param to `numReaders/2`:

```
storage.NewReadLimiter(numReaders/2, nil),
```

frac/active_concurrency_test.go
```diff
@@ -0,0 +1,281 @@
...
+				readNext := func() ([]byte, error) {
+					if idx >= len(bulk) {
+						return nil, nil
+					}
+					d := []byte(bulk[idx])
+					idx++
+					return d, nil
+				}
+
+				proc := indexer.NewProcessor(mapping, tokenizers, 0, 0, 0)
```
eguguchkin: nit: this line can be moved outside the loop (to not create each time)

frac/active_concurrency_test.go
cheb0: fixed

frac/active_concurrency_test.go
cheb0: fixed

frac/active_concurrency_test.go
cheb0: fixed

frac/active_concurrency_test.go
cheb0: fixed

eguguchkin (approved):

PR #246, opened 2025-11-05
refactor(fracmanager): reorganize test structure

### LINKED ISSUE #245: Refactor fracmanager tests
Tasks:
- Move loader tests to a separate file
- Remove old offloading and fetcher tests
- Extract test object creation into shared factory function

eguguchkin:
Fixes #245

---

### DIFF
--- a/fracmanager/loader_test.go
+++ b/fracmanager/loader_test.go
@@ -82,14 +82,15 @@ func TestReplayWithMultipleEmpty(t *testing.T) {
 	// fill data
 	nonEmpty := make([]*common.Info, 0)
 	actives := make([]*frac.Active, 0, fracCount)
-	actives = append(actives, fp.CreateActive())
 	for i := 0; i < fracCount; i++ {
+		active := fp.CreateActive()
 		if i%3 == 0 {
-			appendDocs(t, actives[len(actives)-1], 500+rand.Intn(100))
-			nonEmpty = append(nonEmpty, actives[len(actives)-1].Info())
+			appendDocs(t, active, 500+rand.Intn(100))
+			nonEmpty = append(nonEmpty, active.Info())
 		}
-		actives = append(actives, fp.CreateActive())
+		actives = append(actives, active)
 	}
+	actives = append(actives, fp.CreateActive()) // last active frac is now empty
 
 	// replay and seal
 	active, sealed, err := loader.replayAndSeal(t.Context(), actives)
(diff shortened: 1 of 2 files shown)


tests/integration_tests/offloading_test.go
```diff
@@ -1,173 +0,0 @@
...
-	"github.com/ozontech/seq-db/seq"
-	"github.com/ozontech/seq-db/tests/setup"
-	"github.com/ozontech/seq-db/tests/suites"
-)
-
-type OffloadingSuite struct {
-	suites.Base
-}
-
-func TestOffloading(t *testing.T) {
```
cheb0, line 26: Am I correct that you are deleting this test simply because we now have a unit test which tests search and fetch for remote fractions? My thinking is we can delete this particular test, but we will still need to have some kind of integration test with S3.

eguguchkin: > Am I correct that you are deleting this test simply because we now have a unit test which tests search and fetch for remote fractions?

yes

> My thinking is we can delete this particular test, but we will still need to have some kind of integration test with S3.

We don’t have similar tests yet, apart from this one. However, this test contains a lot of hacks that we wanted to eliminate, which is why work was carried out to test a separately remote fraction.


The fraction offloading logic is partially tested in [this test](https://github.com/ozontech/seq-db/blob/a88cd5b4665c9cf645ca6b8b89d8819fafcf0898/fracmanager/lifecycle_manager_test.go#L119), which is added to the new PR. However, it is not an integration test.

fracmanager/loader_test.go
```diff
@@ -0,0 +1,197 @@
...
+	fp, loader, tearDown := setupLoaderTest(t, &Config{ReplayWorkers: 10})
+	defer tearDown()
+
+	// fill data
+	actives := make([]*frac.Active, 0, fracCount)
+	actives = append(actives, fp.CreateActive())
+	for i := 0; i < fracCount; i++ {
+		appendDocs(t, actives[len(actives)-1], 500+rand.Intn(100))
+		actives = append(actives, fp.CreateActive())
+	}
```
forshev, line 171: nit: some tests here make use of this pattern where docs are added to the last frac in `actives` slice and then new empty frac appended to `actives`, so fracs get filled the next iteration of a loop

maybe change it to more straightforward approach, which i find easier to understand, where new active frac created and filled with docs at the same time, like in `TestReplayWithEmptyActive`:
```go
actives := make([]*frac.Active, 0, fracCount)
	for i := 0; i < fracCount; i++ {
		active := fp.CreateActive()
		appendDocs(t, active, 500+rand.Intn(100))
		actives = append(actives, active)
	}
	actives = append(actives, fp.CreateActive()) // last active frac is now empty
```

what do you think?

forshev (approved):

cheb0 (approved):

Issue #232, opened 2025-10-28
feat: support microsecond resolution for stored documents


cheb0:
Currently, we store docs with millisecond resolution. The goal is to support microsecond resolution, which solves the problem of logs being permuted after the upload if they occurred at the same epoch millisecond.

Main problems/obstacles:
- backward compatibility
- seq-ui integration must be checked
- encoded MIDs block size increase should be estimated

PR #248, opened 2025-11-07
feat: support compatibility with nanosecond MID precision (phase 1) 

cheb0:
# Description

The is the intermediate version for supporting nanosecond MID. This version still operates in millisecond MIDs, but is compatible with the current (main) version as well as with future (phase 2) version where `seq.MID` contains nanoseconds. This allows us to rollback in case of problem with the future version.

- `seq.MID` are still in milliseconds. 
- New frac version to indicate that `seq.MID` are in nanoseconds. This version silently converts them to milliseconds
- New metadata version which contains `seq.MID` in nanoseconds. This version converts them to milliseconds as well.
- New QPR encoding for async version for same reasons.
- Async search info now has version (1 and 2), since it contains `seq.MID` inside stored JSON. If we see a different version (naonseconds), we convert them to millis and change version.
- Frac cache now omits cached info for fractions if fraction version differs from the current version. This is simple because info contains `seq.MID` and we can not change versions.
- Support for new `seq.ID` string representation is added. If MID is in nanoseconds, then delimiter is `_` instead of `-`. In this version, if we see `_` delimiter, then we convert to milliseconds.
- Proxy now can work with new (phase 2) and current stores, i.e. can merge millisecond and nanosecond MID responses. There is a GRPC header which every store sends. This means proxy can operate properly even during the deploy/rollback process. 

Relates to #232 

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/asyncsearcher/async_searcher.go
+++ b/asyncsearcher/async_searcher.go
@@ -32,13 +32,6 @@ import (
 	"github.com/ozontech/seq-db/zstd"
 )
 
-type infoVersion uint8
-
-const (
-	infoVersion1 infoVersion = iota + 1 // MIDs stored in milliseconds
-	infoVersion2                        // MIDs stored in nanoseconds
-)
-
 const (
 	asyncSearchExtInfo      = ".info"
 	asyncSearchExtQPR       = ".qpr"
@@ -47,6 +40,13 @@ const (
 
 	minRetention = 5 * time.Minute
 	maxRetention = 30 * 24 * time.Hour // 30 days
+)
+
+type infoVersion uint8
+
+const (
+	infoVersion1 infoVersion = iota + 1 // MIDs stored in milliseconds
+	infoVersion2                        // MIDs stored in nanoseconds
 )
 
 var (


seq/seq.go
```diff
@@ -83,7 +83,15 @@ func FromString(x string) (ID, error) {
 		return id, err
 	}
 
-	id.MID = MID(binary.LittleEndian.Uint64(mid))
+	delimiter := x[16]
+	if delimiter == '_' {
```
forshev: nit: you can use switch statement here

proxy/search/ingestor.go
```diff
@@ -228,7 +229,18 @@ func (si *Ingestor) singleDocsStream(ctx context.Context, explain bool, source u
...
 
-	var it DocsIterator = newGrpcStreamIterator(stream, host, source, len(ids))
+	md, err := stream.Header()
+	midPrecision := seq.MIDPrecisionMilliseconds
+	if md != nil && err == nil {
+		if precisionValues := md.Get(consts.MIDPrecisionHeader); len(precisionValues) > 0 {
+			midPrecision = seq.ParseMIDPrecision(precisionValues[0])
+		}
+	}
+	if err != nil {
```
forshev, line 235: nit: let's put `err != nil` check right after `md, err := stream.Header()`. it's like usual go way

asyncsearcher/encoding.go
```diff
@@ -40,8 +43,8 @@ func unmarshalQPR(dst *seq.QPR, src []byte, idsLimit int) (_ []byte, err error)
 
 	version := src[0]
 	src = src[1:]
-	if version != qprBinVersion {
-		return nil, fmt.Errorf("invalid QPR version %d; want %d", version, qprBinVersion)
+	if version != qprBinVersion1 && version != qprBinVersion2 {
```
forshev: nit: maybe add map with all the available versions and check presence of a version in this map?

like this:
```go
var availableVersions = map[uint8]struct{}{
	qprBinVersion1: {},
	qprBinVersion2: {},
}
```

and then:
```go
if _, ok := availableVersions[version]; !ok {
	return nil, fmt.Errorf("invalid QPR version %d", version)
}
```

asyncsearcher/async_searcher.go
```diff
@@ -40,6 +40,9 @@ const (
 
 	minRetention = 5 * time.Minute
 	maxRetention = 30 * 24 * time.Hour // 30 days
+
+	infoVersion1 = uint8(1) // MIDs stored in milliseconds
+	infoVersion2 = uint8(2) // MIDs stored in nanoseconds
```
forshev: nit: let's add type alias for infoVersion, something like

```go
type infoVersion uint8

const (
	infoVersion1 infoVersion = iota + 1
	infoVersion2
)
```

asyncsearcher/encoding.go
```diff
@@ -14,10 +14,13 @@ import (
 
 var be = binary.BigEndian
 
-const qprBinVersion = uint8(1)
+const (
+	qprBinVersion1 = uint8(1) // MIDs stored in milliseconds
+	qprBinVersion2 = uint8(2) // MIDs stored in nanoseconds
```
forshev: nit: let's add type alias for qprBinVersion, something like

```go
type qprBinVersion uint8

const (
	qprBinVersion1 qprBinVersion1 = iota + 1
	qprBinVersion2
)
```

indexer/meta_data.go
```diff
@@ -66,7 +66,11 @@ func (m *MetaData) UnmarshalBinary(b []byte) error {
 
 	switch version {
 	case 1:
+		// Version 1 meta stores MID in milliseconds
 		return m.unmarshalVersion1(b)
+	case 2:
+		// Version 2 meta stores MID in nanoseconds
+		return m.unmarshalVersion2(b)
```
forshev, line 73: nit: i don't know which way is better, but we can keep one method for unmarshaling and check versions inside. seems like this approach is more useful when we need to check version somewhere in the middle of a method

```go
switch version {
	case 1, 2:
		return m.unmarshal(b, version)
	default:
		return fmt.Errorf("unimplemented metadata version: %d", version)
	}
```

and then

```go
func (m *MetaData) unmarshal(b []byte) error {
    if version == 1 {
        // some code
    }

    // some code

    if version >= 2 {
        // some code
    }

    // some code
}
```

frac/sealed/seqids/blocks.go
```diff
@@ -26,6 +28,14 @@ func (b *BlockMIDs) Unpack(data []byte) error {
 		return err
 	}
 	b.Values = values
+
+	// v2 (nanosecond MIDs) compatibility - convert nanos to millis
+	if b.fracVersion >= config.BinaryDataV2 {
+		for i := range b.Values {
+			b.Values[i] = uint64(seq.NanosToMID(b.Values[i]))
+		}
+	}
```
forshev, line 36: nit: do we get some performance penalty from additional loop over `values`? if yes, maybe convert mids in one pass inside `unpackRawIDsVarint()`?

```go
func unpackRawIDsVarint(src []byte, dst []uint64, nanoIDs bool) ([]uint64, error)
```

frac/sealed/seqids/loader.go
```diff
@@ -48,7 +48,10 @@ func (l *Loader) GetMIDsBlock(index uint32, buf []uint64) (BlockMIDs, error) {
 		return BlockMIDs{}, err
 	}
 	// unpack
-	block := BlockMIDs{Values: buf}
+	block := BlockMIDs{
+		fracVersion: l.fracVersion,
+		Values:      buf,
+	}
 	if err := block.Unpack(data); err != nil {
```
forshev: nit: seems like this is the only place we need to know fracVersion of a block, maybe pass version as `Unpack` method's parameter instead of a struct's field?

proxy/search/ingestor.go
```diff
@@ -602,15 +614,38 @@ func (si *Ingestor) searchHost(ctx context.Context, req *storeapi.SearchRequest,
...
 		grpc.MaxCallSendMsgSize(256*int(units.MiB)),
 		grpc.UseCompressor(gzip.Name),
+		grpc.Header(&md),
 	)
 	if err != nil {
 		return nil, 0, err
 	}
 
+	// Check the store's MID precision from response header
+	// If header indicates milliseconds, then convert to nanoseconds
```
forshev: `If header indicates milliseconds, then convert to nanoseconds` -> `If header indicates nanoseconds, then convert to milliseconds`?

proxy/search/ingestor.go
```diff
@@ -602,15 +614,38 @@ func (si *Ingestor) searchHost(ctx context.Context, req *storeapi.SearchRequest,
...
+		}
+
+		if len(data.Histogram) > 0 {
+			newHist := make(map[uint64]uint64, len(data.Histogram))
+			for mid, v := range data.Histogram {
+				newHist[uint64(seq.NanosToMID(mid))] = v
+			}
+			data.Histogram = newHist
+		}
+	}
```
forshev, line 657: what about aggs with timeseries here? afair bucket keys contain MID

proxy/search/async.go
```diff
@@ -154,13 +157,34 @@ func (si *Ingestor) FetchAsyncSearchResult(
...
+					}
+
+					if len(response.Histogram) > 0 {
+						newHist := make(map[uint64]uint64, len(response.Histogram))
+						for mid, v := range response.Histogram {
+							newHist[uint64(seq.NanosToMID(mid))] = v
+						}
+						response.Histogram = newHist
+					}
+				}
```
forshev, line 189: what about aggs with timeseries here? afair bucket keys contain MID

asyncsearcher/async_searcher.go
cheb0: done

asyncsearcher/encoding.go
cheb0: done

asyncsearcher/encoding.go
cheb0: done

frac/sealed/seqids/blocks.go
cheb0: good point. I'm pretty sure we have performance penalty. However, this condition

`if b.fracVersion >= config.BinaryDataV2 {`

will be true only when we rollback from phase 2 to phase 1 and search phase 2 fractions (there will be relatively small amount of data). This is a relatively unlikely event.

However, we will have similar code block here in phase 2 (which will replace this code block) where we convert old fraction MIDs to nanoseconds. Since we have tons of old fractions, we will need to convert it in a single pass. Another problem is `unpackRawIDsVarint` is used by old version RID blocks, so I guess we will need to copy this method.

frac/sealed/seqids/loader.go
cheb0: done

indexer/meta_data.go
cheb0: I dunno, looks like it all the same. I'd rather prefer dedicated methods/unmarshallers for every version.

proxy/search/ingestor.go
cheb0: fixed

proxy/search/ingestor.go

seq/seq.go
cheb0: done

seq/seq.go
```diff
@@ -83,7 +83,15 @@ func FromString(x string) (ID, error) {
 		return id, err
 	}
 
-	id.MID = MID(binary.LittleEndian.Uint64(mid))
+	switch delimiter := x[16]; delimiter {
+	case '_':
```
eguguchkin, line 93: nit: let's take another look at the different delimiter options for the new format?

```
/messages/de4f04a69a010000-e80225a3b2658e8c // old
/messages/de4f04a69a010000_e80225a3b2658e8c // option 1 with _
/messages/de4f04a69a010000:e80225a3b2658e8c // option 2 with :
/messages/de4f04a69a010000.e80225a3b2658e8c // option 3 with .
```
@forshev @dkharms @moflotas tell me, please, how important is it for you that a specific symbol is used in the new format? If so, which option?

it will be difficult to change later

indexer/meta_data.go
eguguchkin: nit: this looks good for me.

But i don't like name of method unmarshalVersion1And2 :)

And I will do something like this:

```go

func (m *MetaData) unmarshalVersion1(b []byte) error {
	return m.decode(b)
}

func (m *MetaData) unmarshalVersion2(b []byte) error {
	if err := m.decode(b); err != nil {
		return err
	}
	m.ID.MID = seq.NanosToMID(uint64(m.ID.MID))
	return nil
}

func (m *MetaData) decode(b []byte) error {
	m.ID.MID = seq.MID(binary.LittleEndian.Uint64(b))
	b = b[8:]
	m.ID.RID = seq.RID(binary.LittleEndian.Uint64(b))
	b = b[8:]

	// Decode uncompressed document size.
	m.Size = binary.LittleEndian.Uint32(b)
	b = b[4:]

	// Decode tokens length.
	toksLen := binary.LittleEndian.Uint32(b)
	b = b[4:]

	// Decode tokens.
	m.Tokens = m.Tokens[:0]
	m.Tokens = slices.Grow(m.Tokens, int(toksLen))
	var err error
	for i := uint32(0); i < toksLen; i++ {
		var token tokenizer.MetaToken
		b, err = token.UnmarshalBinary(b)
		if err != nil {
			return err
		}
		m.Tokens = append(m.Tokens, token)
	}
	return nil
}

```

network/grpcutil/interceptors.go
```diff
@@ -123,3 +125,35 @@ func PassMetadataUnaryClientInterceptor() grpc.UnaryClientInterceptor {
...
+}
+
+// MIDPrecisionHeaderStreamServerInterceptor sets the MID precision header for all streaming responses.
+func MIDPrecisionHeaderStreamServerInterceptor(precision seq.MIDPrecision) grpc.StreamServerInterceptor {
+	return func(
+		srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo,
+		h grpc.StreamHandler,
+	) error {
+		md := metadata.New(map[string]string{
+			consts.MIDPrecisionHeader: precision.String(),
```
eguguchkin: nit: сurrently, the header name is self-explanatory and describes its content well, which is convenient. However, when we scale the system, we might end up with many such unique headers, which would complicate their processing.

What do you think about using a more universal name, for example, X-Seq-Protocol-ID, and passing the protocol version number in it?

proxy/search/ingestor.go
cheb0: Timestamps in aggregation bins are transferred as protobuf timestamps between store and proxy. hence,  It's not a problem. The problem arises only when we either:
- store MID on disk
- transfer raw MID as `uint64` in protobuf between store and proxy. We can't differentiate if this nano or milli

Just to clarify, I have a test which tests deploy and rollback for cluster setup: two stores and one proxy.
- send 500k docs (all phase 1)
- validate agg query (count with interval `1s`)
- replace one store with phase 2 ver
- send another 500k docs
- validate agg query for 1M docs
- replace phase 2 store with phase 1 (rollback)
- send another 500k and validate results

Agg result fully matches with expected (in-memory computed)

proxy/search/async.go
cheb0: same as for ordinary (explained below) agg. for async search we have a QPR encoding support for MIDs in agg results and I also have tests for async search aggregation

seq/seq.go
forshev: underscore looks good to me

network/grpcutil/interceptors.go
cheb0: Switched to protocol. I named the header `x-seq-protocol-id` without capital letters simply because there is already another header `DebugHeader = "x-o3-sample-trace"` named in that way.

PR #250, opened 2025-11-07
fix: add distinct aggregation limit error messages

### LINKED ISSUE #249: Return clear error messages when aggregation limit exceeded
There are three distinct limits for aggregations, but we return the same error message for each of them

Need to make distinct error messages so response gives full information about which specific limit was exceeded

forshev:
# Description

Added separate error message for each of aggregation limits

Fixes #249

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```


proxy/search/ingestor.go
```diff
@@ -567,8 +567,14 @@ func (si *Ingestor) searchShard(
...
-			if errMessage == consts.ErrTooManyUniqValues.Error() {
-				return nil, source, fmt.Errorf("store forbids aggregation request: %w", consts.ErrTooManyUniqValues)
+			if errMessage == consts.ErrTooManyFieldTokens.Error() {
+				return nil, source, fmt.Errorf("store forbids aggregation request: %w", consts.ErrTooManyFieldTokens)
+			}
+			if errMessage == consts.ErrTooManyGroupTokens.Error() {
+				return nil, source, fmt.Errorf("store forbids aggregation request: %w", consts.ErrTooManyGroupTokens)
+			}
+			if errMessage == consts.ErrTooManyFractionTokens.Error() {
+				return nil, source, fmt.Errorf("store forbids aggregation request: %w", consts.ErrTooManyFractionTokens)
```
moflotas, line 577: I believe this is old behaviour, which is no currently used (since, we migrated from string comparisons to protobuf error-codes) and we should remove this code

frac/processor/aggregator_test.go
```diff
@@ -251,3 +252,37 @@ func TestSingleTreeCountAggregator(t *testing.T) {
 		r.Equal(hist.Total, result.SamplesByBin[token].Total)
 	}
 }
+
+func TestAggregatorLimitExceeded(t *testing.T) {
+	// For now input for this test is incorrect since we support
+	// aggregations only for `keyword` index type.
+	// Will be fixed in #310.
```
moflotas, line 259: nit: maybe add `TODO` to not forget about removing this code

moflotas (changes requested):

proxy/search/ingestor.go
moflotas: discussed internally and decided to make those changes separately

moflotas (approved):

PR #256, opened 2025-11-10
fix: do not add system MID to fraction distribution

cheb0:
# Description

- skip system MID while building MID distribution
- system MID is `MaxUint64` which always went to `0` index bucket (bit) in bitmask
- fix MID->Time conversion for `MaxUint64`. Previously, `1970` year was derived from `MaxUint64` MID. Integration tests use `0` as `from` MID and `MaxUint64` as `to` MID. Therefore, both `0` and `MaxUint64` MIDs convert to bucket (bit) 0 due to overflow, i.e. calling `IsIntersecting(0, MaxUint64)` is essentially checking that 0th bit is set in distribution. Therefore overflow must also be fixed in this PR.

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```


seq/seq.go
```diff
@@ -19,7 +20,14 @@ type RID uint64 // random part of ID
 type LID uint32 // local id for a fraction
 
 func (m MID) Time() time.Time {
-	return time.UnixMilli(int64(m))
+	if uint64(m) <= math.MaxInt64 {
+		return time.UnixMilli(int64(m))
+	} else {
+		millis := uint64(m)
+		secondsPart := millis / 1000
+		nanosPart := millis - secondsPart*1000
```
eguguchkin: Same thing:

```diff
- nanosPart := millis - secondsPart*1000
+ nanosPart := millis % 1000
```

But are these really nanoseconds? I think it's the millisecond part. But if we convert it to nanoseconds, we'll get the exactly `time.UnixMilli()` function with overflow.

and I don't know how to solve this problem fairly

eguguchkin: Perhaps a suitable solution would be to simply truncate the excessively large MID value, so that anything greater than math.MaxUint64 / 2 is rounded to math.MaxUint64 / 2?

In millisecond terms, this is the year 292278994 AD.

eguguchkin: Your solution has almost the same behavior, but with some magic that is incomprehensible to the reader.

cheb0: I think when I was implementing it I was thinking about not doing any calculations at all, if there is no overflow. i.e. it's a fast path: if millis fit in int64 positive values range, then just cast MID  to int64 directly and do not do anything else. 

Right now, I changed else branch not to do any calculations at all, since it's far in the future.

At this point, it's enough not to overflow. In phase 2 branch (nanos MID) this method will be fully replaced with another one.

PR #266, opened 2025-11-18
feat(fracmanager): add handling for upload queue overflow and disk space exhaustion

### LINKED ISSUE #265: Add handling for upload queue overflow and disk space exhaustion
**Description:**
Need to implement handling for two overflow scenarios:

**Offloading queue overflow**

When the upload queue reaches maximum capacity

- Automatically remove the oldest uploadable fractions
- Ensure proper priority handling for upload operations

**Disk space exhaustion**

When insufficient disk space for new data writes

- Reject write operations
- Switch the active fraction to "suspended" mode

eguguchkin:
Fixes #265

---

### DIFF
--- a/fracmanager/fraction_registry.go
+++ b/fracmanager/fraction_registry.go
@@ -6,10 +6,11 @@ import (
 	"sync"
 	"time"
 
+	"go.uber.org/zap"
+
 	"github.com/ozontech/seq-db/frac"
 	"github.com/ozontech/seq-db/logger"
 	"github.com/ozontech/seq-db/util"
-	"go.uber.org/zap"
 )
 
 // fractionRegistry manages fraction queues at different lifecycle stages.


fracmanager/fraction_registry.go
```diff
@@ -151,6 +153,31 @@ func (r *fractionRegistry) RotateIfFull(maxSize uint64, newActive func() *active
 	return old, wg.Wait, nil
 }
 
+func (r *fractionRegistry) SuspendIfOverCapacity(maxQueue, maxSize uint64) bool {
```
forshev: nit: seems like return value is never used

config/config.go
dkharms, line 1: Please add corresponding validations (of the new config options) into `config/validation.go`.

fracmanager/lifecycle_manager.go
```diff
@@ -36,18 +37,26 @@ func newLifecycleManager(
...
 }
 
 // Maintain performs periodic lifecycle management tasks.
 // It coordinates rotation, offloading, cleanup based on configuration.
-func (lc *lifecycleManager) Maintain(ctx context.Context, config *Config, wg *sync.WaitGroup) {
-	lc.rotate(config.FracSize, wg)
-	if config.OffloadingEnabled {
-		lc.offloadLocal(ctx, config.TotalSize, wg)
-		lc.cleanRemote(config.OffloadingRetention, wg)
+func (lc *lifecycleManager) Maintain(ctx context.Context, cfg *Config, wg *sync.WaitGroup) {
```
dkharms, line 46: nit: Maybe you can add method `Maintain` to `FracManager` as well? It will grab mutexes and call `lifecycleManager.Maintain`.

Now it's done like this in `startMaintWorker`:
```go
fm.mu.Lock()
// Perform fraction maintenance (rotation, truncating, offloading, etc.)
fm.lc.Maintain(ctx, cfg, wg)
fm.mu.Unlock()
```

I do not like that we grab mutexes somewhere in the related code. So I suggest to do just:
```go
// Perform fraction maintenance (rotation, truncating, offloading, etc.)
fm.Maintain(ctx, cfg, wg)
```

fracmanager/lifecycle_manager.go
```diff
@@ -36,18 +37,26 @@ func newLifecycleManager(
...
+
```
dkharms: nit: Empty line.

fracmanager/lifecycle_manager.go
```diff
@@ -36,18 +37,26 @@ func newLifecycleManager(
...
+	suspendThreshold := cfg.TotalSize + cfg.TotalSize/100 + cfg.OffloadingQueueSize
+	lc.registry.SuspendIfOverCapacity(cfg.SealingQueueLen, suspendThreshold)
+
+	lc.rotate(cfg.FracSize, wg)
+	if cfg.OffloadingEnabled {
+		lc.offloadLocal(ctx, cfg.TotalSize, cfg.OffloadingRetryDelay, wg)
+		if cfg.OffloadingQueueSize > 0 {
+			lc.removeOverflowed(cfg.OffloadingQueueSize, wg)
+		}
```
dkharms, line 54: Shouldn't we check for overflow *before* making an attempt to offload sealed fractions?

fracmanager/tasks.go
```diff
@@ -0,0 +1,67 @@
...
+
+func NewTaskManager() *TaskManager {
+	return &TaskManager{
+		running: make(map[string]*task),
+	}
+}
+
+// Run starts a new background task with the given ID and context.
+// The task will be automatically removed when completed.
+func (t *TaskManager) Run(id string, ctx context.Context, action func(ctx context.Context)) *task {
```
dkharms: What about task deduplication? Does caller guarantee that it will not provide task with the same id several times?

fracmanager/lifecycle_manager.go
```diff
@@ -136,7 +146,41 @@ func (lc *lifecycleManager) offloadLocal(ctx context.Context, sizeLimit uint64,
...
 			sealed.instance.Suicide()
 			maintenanceTruncateTotal.Add(1)
-		}()
+		})
+	}
+}
+
+// OffloadWithRetry attempts to offload a fraction with retries until success or cancellation.
+// Returns the remote fraction instance and a boolean indicating whether offloading was not canceled.
+func (lc *lifecycleManager) offloadWithRetry(ctx context.Context, sealed *frac.Sealed, retryDelay time.Duration) *frac.Remote {
```
dkharms, line 156: Previously we attempted to offload fraction `Offloading.RetryCount` times and if there is no success we delete the fraction.

Now, there might be more than `Offloading.RetryCount` attempts.

Also `Offloading.RetryDelay` is ambiguous. When I saw this option I though that it is duration between S3 client retries. Not the duration between retry groups (let's call it this way).

fracmanager/lifecycle_manager.go
eguguchkin: I think it should be the other way around — if we remove the overflowed fractions first and bring the queue size down to the limit, we'll then add new ones back, and until the next maintenance run, the queue will exceed the limit again.

fracmanager/fraction_registry.go
eguguchkin: fixed: removed returned value

config/config.go
eguguchkin: done

fracmanager/lifecycle_manager.go
eguguchkin: done

fracmanager/lifecycle_manager.go
eguguchkin: done

fracmanager/tasks.go
eguguchkin: fixed

fracmanager/lifecycle_manager.go
eguguchkin: As discussed, the `RetryCount` on the S3 client has been removed. Retries are now managed by the FracManager, which runs the upload in a loop until the resources are exhausted. This eliminates the ambiguity around `RetryDelay` and unifies the logic.

PR #272, opened 2025-11-20
perf: inline `binary.Varint` by hand

dkharms:
# Description

I was running our benchmarks suite and I noticed that for full-text-search scenarios we waste a lot of CPU on `binary.Varint` when doing `seq.MID` unpacking in a hot loop (when search request wants data that was not previously cached).

So I decided to inline that function by hand and ran following search scenario:
```bash
k6 run \
	--summary-trend-stats 'avg,min,med,max,p(90),p(95),p(99)' \
	-e BASE_URL="http://localhost:9002" \
	seq-db-fetch-5k-fulltext.js
```

*Each run took around 60s. I've made two runs (for baseline and modified versions, so four in total) and took results from the second run for each version.*

Before:
```text
HTTP
http_req_duration..............: avg=111.93ms min=3.05ms   med=99.8ms   max=324.25ms p(90)=212.2ms  p(95)=229.13ms p(99)=300.88ms
  { expected_response:true }...: avg=111.93ms min=3.05ms   med=99.8ms   max=324.25ms p(90)=212.2ms  p(95)=229.13ms p(99)=300.88ms
http_req_failed................: 0.00%  0 out of 3847
http_reqs......................: 3847   63.808591/s
```

After:
```text
HTTP
http_req_duration..............: avg=99.81ms min=2.76ms   med=94.59ms  max=295.87ms p(90)=166.47ms p(95)=182.62ms p(99)=230.39ms
  { expected_response:true }...: avg=99.81ms min=2.76ms   med=94.59ms  max=295.87ms p(90)=166.47ms p(95)=182.62ms p(99)=230.39ms
http_req_failed................: 0.00%  0 out of 4004
http_reqs......................: 4004   66.543389/s
```

Also I've noticed improvements in CPU usage (the latter two hills are from modified version, 2 CPU vs 2.35 CPU):
image

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/frac/sealed/seqids/blocks.go
+++ b/frac/sealed/seqids/blocks.go
@@ -80,17 +80,16 @@ func unpackRawIDsVarint(src []byte, dst []uint64) ([]uint64, error) {
 	dst = dst[:0]
 	id := uint64(0)
 	for len(src) != 0 {
-		ux, n := binary.Uvarint(src) // ok to continue in presence of error
+		udelta, n := binary.Uvarint(src)
 		if n <= 0 {
 			return nil, errors.New("varint decoded with error")
 		}
 
-		x := int64(ux >> 1)
-		if ux&1 != 0 {
-			x = ^x
+		delta := int64(udelta >> 1)
+		if udelta&1 != 0 {
+			delta = ^delta
 		}
 
-		delta := x
 		id += uint64(delta)
 		dst = append(dst, id)
 


frac/sealed/seqids/blocks.go
```diff
@@ -80,13 +80,21 @@ func unpackRawIDsVarint(src []byte, dst []uint64) ([]uint64, error) {
 	dst = dst[:0]
 	id := uint64(0)
 	for len(src) != 0 {
-		delta, n := binary.Varint(src)
+		ux, n := binary.Uvarint(src) // ok to continue in presence of error
```
dkharms: Come up with better naming.

dkharms:
@cheb0 If you could validate these result that would be very helpful.

cheb0:
> @cheb0 If you could validate these result that would be very helpful.

I decided not to go with current search benches, but to check how it aligns with more realistic search request performance measurements. I have ~400 Mb of data locally and store+proxy set up.

I had three runs of 10 iterations each. On every iteration, we restart store and proxy and then issue a query twice to check how cache affects the query performance. And then we compute avg time in millis. Here is the results

**query**: k8s_pod:payment-backend-* AND level:4, offset=2000, size=6000
_Before (main)_
Avg: 217.2 ms
Avg (cached): 64.9 ms
Avg: 217.6 ms
Avg (cached): 67.6 ms
Avg: 215.7 ms
Avg (cached): 66.7 ms

_After (PR)_
Avg: 215.5 ms
Avg (cached): 64.5 ms
Avg: 213.2 ms
Avg (cached): 64.7 ms
Avg: 217.8 ms
Avg (cached): 65.8 ms

**query**: k8s_pod:payment-backend-0 AND level:1 AND message:database AND NOT message:optimization AND NOT message:transaction, offset=2000, size=6000
_Before (main)_
Avg: 291.1 ms
Avg (cached): 214.6 ms
Avg: 291.9 ms
Avg (cached): 219.0 ms
Avg: 287.9 ms
Avg (cached): 215.0 ms

_After (PR)_
Avg: 285.6 ms
Avg (cached): 213.4 ms
Avg: 288.6 ms
Avg (cached): 215.5 ms
Avg: 287.8 ms
Avg (cached): 214.5 ms

I think there is clearly a difference (this PR is faster) but not so drastic as when search requests are issued non-stop by multiple threads. In any case, it's just a tip of the iceberg in terms of search benchmarking. 

UPD: I made the query more complex and reduced size to 100, because fetching docs took most time for the previous test. Now numbers are better

_Before (main)_
Avg: 74.0 ms
Avg (cached): 8.3 ms
Avg: 70.3 ms
Avg (cached): 8.2 ms
Avg: 71.5 ms
Avg (cached): 8.3 ms

_After (PR)_
Avg: 76.2 ms
Avg (cached): 7.5 ms
Avg: 73.6 ms
Avg (cached): 7.5 ms
Avg: 70.1 ms
Avg (cached): 7.2 ms

PR #277, opened 2025-11-24
feat(fracmanager): implement fraction snapshots with wait group reference counting

### LINKED ISSUE #275: Refactoring Fraction Lifecycle Management System
### Problem
The current fraction management implementation uses a proxy wrapper with mutex for safe access to the changing list of fractions. This approach does not scale for planned compaction and partitioning features where:
- Multiple fractions can merge into one
- One fraction can split into several
- Fraction lifecycle becomes more complex and doesn't fit the current model

### Task
Replace the current proxy wrapper system with a fraction snapshots mechanism using reference counting via `wait group`.

### Proposed Solution
1. **Fraction Snapshots**: Create a snapshot of the fraction list whenever its composition changes
2. **Reference Counting**: Use `wait group` to track active uses of fractions
3. **Safe Deletion**: Guarantee that a fraction won't be deleted while it's being used by someone

eguguchkin:
Fixes #275

---


storeapi/grpc_search.go
```diff
@@ -164,13 +164,16 @@ func (g *GrpcV1) doSearch(
...
 	searchTr := tr.NewChild("search iteratively")
+	fractions, release := g.fracManager.FractionsSnapshot()
 	qpr, err := g.searchData.searcher.SearchDocs(
 		ctx,
-		g.fracManager.Fractions(),
+		fractions,
 		searchParams,
 		tr,
 	)
+	release()
```
cheb0: If search panics, then `release()` is never called.

fracmanager/fractions_snapshot.go
```diff
@@ -0,0 +1,93 @@
...
+		len(remotes) +
+		len(offloading) +
+		len(locals) +
+		len(sealing)
+
+	fs := fractionsSnapshot{fractions: make([]frac.Fraction, 0, capacity)}
+
+	// Collect fractions in correct chronological order: from oldest (remote) to newest (active)
+	for _, r := range remotes {
+		r.wg = &fs.wg
```
cheb0: What happens in the following scenario:
- we already have a snapshot built - S1
- search request arrives
- while search request is being executed, snapshot rebuild is called. Let's call it S2
- now we overwriting a wait group inside `...SyncDestroyable` struct with S2 wait group.
- at the same time, if we want to destroy a fraction, we call `wg.Wait()`. But we might issue wait for wait group S2. Since nobody has called `wg.Add` for a S2 wg, it will return immediately. However, search request might still be working and touch this fraction.

cheb0: Besides, we issue `s.wg = &fs.wg` under fraction registry lock (only a single snapshot can be being under construction at the same time). 
However, we read this wg group `d.wg.Wait()` in `Destroy`. 

i.e. techically it's a data race if understand correctly - it's a unsynchronized read/write of a non-atomic variable. The whole question is what guarantees do we have in Go.

fracmanager/fraction_registry.go
```diff
@@ -18,18 +18,18 @@ type fractionRegistry struct {
...
 	stats       registryStats // size statistics for monitoring
 	oldestTotal uint64        // creation time of oldest fraction in all list including remote
 	oldestLocal uint64        // creation time of oldest fraction in local or offloading queues
 
-	muAll  sync.RWMutex    // protects active, all, and oldestTotal fields
-	active *activeProxy    // currently active writable fraction
-	all    []frac.Fraction // all fractions in creation order (read-only view)
+	muAll    sync.RWMutex       // protects active, all, and oldestTotal fields
+	appender *syncAppender      // currently active writable fraction
+	all      *fractionsSnapshot // all fractions in creation order (read-only view)
```
cheb0: If we have search request constantly arriving (let's say 5 requests per second) and taking same fraction snapshot. Can we potentially run into any problems?

fracmanager/fractions_snapshot.go
eguguchkin: Yup, the logic was broken here — I’ve rewritten it completely, one more time. :)

fracmanager/fracs_stats.go
```diff
@@ -95,3 +95,7 @@ func (s *registryStats) SetMetrics() {
 	s.offloading.SetMetrics(dataSizeTotal, "offloading")
 	s.remotes.SetMetrics(dataSizeTotal, "remotes")
 }
+
+func (s registryStats) TotalSizeOnDiskLocal() uint64 {
+	return s.sealing.totalSizeOnDisk + s.sealed.totalSizeOnDisk
```
dkharms, line 100: nit: Why `Active` fraction is not accounted here?

util/min_heap.go
```diff
@@ -0,0 +1,117 @@
+package util
+
+import (
+	"cmp"
+	"container/heap"
+)
+
+// MinHeap is a min‑heap for any comparable type.
+// Maintains both a heap structure and a map for fast lookup of items.
+type MinHeap[T cmp.Ordered] struct {
```
dkharms, line 10: Maybe we can relax this a little bit to `MinHeap[T comparable]`?

util/min_heap.go
```diff
@@ -0,0 +1,117 @@
...
+
+// Swap is part of heap.Interface — swaps elements and updates their indices.
+func (hw *heapWrapper[T]) Swap(i, j int) {
+	hw.items[i], hw.items[j] = hw.items[j], hw.items[i]
+	hw.items[i].index = i
+	hw.items[j].index = j
+}
+
+// Push is part of heap.Interface — adds a new element to the heap.
+func (hw *heapWrapper[T]) Push(x interface{}) {
```
dkharms:
```suggestion
func (hw *heapWrapper[T]) Push(x any) {
```

util/min_heap.go
```diff
@@ -0,0 +1,117 @@
...
+	item := x.(*heapItem[T])
+	item.index = len(hw.items)
+	hw.items = append(hw.items, item)
+}
+
+// Pop is part of heap.Interface — removes and returns the last element.
+func (hw *heapWrapper[T]) Pop() interface{} {
```
dkharms:
```suggestion
func (hw *heapWrapper[T]) Pop() any {
```

storeapi/grpc_v1.go
```diff
@@ -98,6 +98,9 @@ type GrpcV1 struct {
 }
 
 func NewGrpcV1(cfg APIConfig, fracManager *fracmanager.FracManager, mappingProvider MappingProvider) *GrpcV1 {
+	fractions, release := fracManager.Fractions()
+	defer release()
```
dkharms: Are you sure that this is a correct place to call `release()`? All fractions will be released when function returns however fractions are claimed by `asyncsearcher`.

storeapi/grpc_async_search.go
```diff
@@ -46,8 +46,12 @@ func (g *GrpcV1) StartAsyncSearch(
 		Retention: r.Retention.AsDuration(),
 		WithDocs:  r.WithDocs,
 	}
-	fracs := g.fracManager.Fractions().FilterInRange(seq.MillisToMID(uint64(r.From)), seq.MillisToMID(uint64(r.To)))
-	if err := g.asyncSearcher.StartSearch(req, fracs); err != nil {
+
+	fracs, release := g.fracManager.Fractions()
+	defer release()
```
dkharms: And the same question here (https://github.com/ozontech/seq-db/pull/277/changes#r3233437945)

fracmanager/sync_appender.go
```diff
@@ -0,0 +1,82 @@
...
+	a.mu.RLock()
+	if a.finalized {
+		a.mu.RUnlock()
+		return ErrFractionNotWritable
+	}
+	if a.suspended {
+		a.mu.RUnlock()
+		return ErrFractionSuspended
+	}
+	a.wg.Add(1) // Important: wg.Add() inside lock to prevent race with WaitWriteIdle()
```
dkharms, line 40: Is it really necessary? I mean this comment:
```
 // Important: wg.Add() inside lock to prevent race with WaitWriteIdle()
```

`WaitWrideIdle()` is always called after `Finalize()` or `Suspend()` -- there is happens-before relationship between `Append()` and those two methods because the are serialized against the same mutex.

So, it is not possible to squeeze another append between `Finalize()` and `WaitWriteIdle()`. And even more -- `WaitWriteIdle()` does not take locks.

fracmanager/sync_appender.go
```diff
@@ -0,0 +1,82 @@
...
+// WaitWriteIdle waits for all pending write operations to complete
+// Used before sealing to ensure data consistency.
+func (a *syncAppender) WaitWriteIdle() {
+	start := time.Now()
+	logger.Info("waiting fraction to stop write...", zap.String("name", a.BaseFileName))
+	a.wg.Wait()
+	waitTime := util.DurationToUnit(time.Since(start), "s")
+	logger.Info("write is stopped",
+		zap.String("name", a.BaseFileName),
+		zap.Float64("time_wait_s", waitTime))
```
dkharms:
```suggestion
	logger.Info(
		"write is stopped",
		zap.String("name", a.BaseFileName),
		zap.Float64("time_wait_s", waitTime),
	)
```

dkharms (approved):

fracmanager/fracs_stats.go
eguguchkin: Because in this case we get data from the stats object (recalculated once a minute after faction rotation), but the active faction changes every second and is not in the stats at all.

Therefore, this function is then used like this:

```go
	// calculate total used disk space
	totalUsedSize := r.stats.TotalSizeOnDiskLocal() + r.appender.Info().FullSize()
```

util/min_heap.go
eguguchkin: That's what I wanted to do from the beginning, but it doesn't work that way. Strange as it may seem.

image

storeapi/grpc_v1.go
eguguchkin: I've already reworked this part a bit. But the previous version was working: the idea is that the list of factions is passed to the method and used to create a persistent queue for processing. Then the necessary BACKGROUND processes are started and the function completes. That is, we only hold the lock while filling the queue, not while processing it.

fracmanager/sync_appender.go
eguguchkin: Here's what happens if you move Add out of the lock

image

fracmanager/sync_appender.go
eguguchkin: done

util/min_heap.go
eguguchkin: done

util/min_heap.go
eguguchkin: done

fracmanager/partitioned_collection.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+}
+
+// Len returns the number of objects in the collection.
+func (c *PartitionedCollection[T]) Len() int {
+	return len(c.byKey)
+}
+
+// GetAllPartitions returns a map of all partitions in the collection.
+func (c *PartitionedCollection[T]) GetAllPartitions() map[uint64][]T {
```
cheb0: nit: seems it's not used anywhere

fracmanager/partitioned_collection.go
```diff
@@ -0,0 +1,117 @@
...
+		c.minPartition.Push(partitionID)
+		c.byPartition[partitionID] = make(map[string]T)
+	}
+	c.byPartition[partitionID][key] = obj
+	c.byKey[key] = obj
+}
+
+// Delete removes an object from the collection by its key.
+// Does nothing if the key doesn't exist.
+func (c *PartitionedCollection[T]) Del(key string) {
```
cheb0, line 45: nit: comment doesn't match function name

storeapi/grpc_search.go
```diff
@@ -229,6 +224,13 @@ func (g *GrpcV1) doSearch(
 	return buildSearchResponse(qpr), nil
 }
 
+func (g *GrpcV1) searchDocs(ctx context.Context, sp processor.SearchParams, tr *querytracer.Tracer) (*seq.QPR, error) {
+	fracs, release := g.fracManager.AcquireFractions()
```
cheb0, line 228: it means we increment a ref count on every fraction for every search request, right?

cheb0 (approved):

storeapi/grpc_search.go
dkharms: @eguguchkin Yeah, maybe we should move `prepareFracs` to `FracManager`?
This way we would increment reference count only for fractions that will be touched within request.

eguguchkin: Actually, yes. We inc counters for EACH fracs

eguguchkin: This could potentially block faction deletion for several tens of seconds and also risks overflowing the allocated disk space. However, in our current environment, this seems unlikely and not critical.

Nevertheless, looking ahead, we would like to be able to run long analytical queries (5–20 minutes), and in that scenario, the deletion delay could become critical. So I think it makes sense to revisit the locking approach later, after implementing timestamp binning.

PR #281, opened 2025-12-01
chore: migrate metrics to OpenMetrics standard

### LINKED ISSUE #280: Migrate metrics to OpenMetrics standard
[OpenMetrics standard](https://github.com/prometheus/OpenMetrics/blob/main/specification/tOpenMetrics.md) specifies naming conventions for metfics. We need to update our metric names to match the standard.

It's also recommended to have _help_ defined for every metric

cheb0:
# Description

Migrated metrics to OpenMetrics standard. I will list the primary rules of the standard and what has been changed.

- Counter metrics **must** have `_total` suffix in their name. I added `_total` suffix to all counters where it's necessary.
- All other metrics (except for counters) **must not** have any of the following suffixes: `_total`, `_count`, `_created`, `_sum`, `_bucket`. I removed `_total` suffix from all gauge/hist metrics. There was a single metric with `_count` suffix, which was also renamed.
- It's **recommended** to have unit types at the end of the metric if appropriate: seconds, bytes, fractions, etc. I added _seconds or  _bytes where appropriate. For oldest entry `timestamp_seconds` are units which are recommended to use. There is prometheus naming best practices: https://prometheus.io/docs/practices/naming/
- It's **recomended** to avoid short names and abbreviations for unit types like `sec` instead of `seconds`. I changed `_sec` to `_seconds` whenever I could.
- It's **recommended** to avoid short terms if they are not commonly used. I replaced a few places like `tracer_fraction_reg_search_sec` -> `tracer_fraction_regular_search_seconds` in the metrics names, but cases like `agg`, `docs` and `meta` are used commonly and could be understood easily. I left them as it is. I also unified `documents` and `docs` in metric names - now `docs` term is used everywhere.
- It's **not recommended** to have metric type in the name. We do not have such metrics. Some metrics have `hist` name in them, but it's acceptable since they do not describe metric type but rather an operation type (search with hist param).
- It's **recommended** to have help for each metric. I added help for all metrics, let's review if help makes sense.
- snake_case is **recommended** for metric names and labels. Already done.
- `le`  и `quantile` are **forbidden** names for labels. We didn't have any of them.

I also found the following cases:
- metrics `TokenizerTokensPerMessage`, `TokenizerParseDurationSeconds` are not used currently
- `skippedIndexes`, `skippedIndexesBytes` (used for tokenization) are currently located in metric/store.go, i.e. place where store metrics. perhaps we could move them to `ingestor` or `tokenizer` metrics.go file
- `skip_shard_total` - seems like an inconsistent name since we choose a shard for storing but loop through replicas and skip a replica if it was already chosen for storing the current bulk.

**Suggestion**: we should have a development guidlines file where we could store rules like "counters must end with _total" and "metrics should never end with _total"

Fixes #280 

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

## Summary by CodeRabbit

* **Documentation**
  * Enhanced Prometheus metrics documentation with clearer help text across multiple modules, improving observability and monitoring clarity.
  * Standardized metric naming conventions with more explicit suffixes (_total, _bytes, _seconds) for better metric identification and consistency in monitoring systems.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,6 +1,6 @@
 # yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
 
-name: Continuous Benchmarks
+name: Benchmarks
 
 on:
   push:
@@ -10,10 +10,25 @@ on:
 
 permissions:
   contents: write
-  deployments: write
 
 jobs:
-  ci-benchmarks:
+  ci-continuous-benchmarks:
+    if: ${{ github.ref_name == 'main' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Deploy latest (unstable) version of seq-db
+        run: |
+          curl --insecure -X POST "https://${{ secrets.CONTINUOUS_BENCHMARKS_ENDPOINT }}/seqagent/v1/b/up" \
+            -H "Content-Type: application/json" \
+            -H "X-Seqbench-Secret: ${{ secrets.SEQBENCH_SECRET }}" \
+            -d @- <<EOF
+          {
+            "refname": "${{ github.ref_name }}",
+            "commit_sha": "${{ github.sha }}"
+          }
+          EOF
+
+  ci-micro-benchmarks:
     runs-on: self-hosted
     steps:
       - name: Checkout repository
--- a/storeapi/grpc_search.go
+++ b/storeapi/grpc_search.go
@@ -83,7 +83,7 @@ func (g *GrpcV1) doSearch(
 
 	// in store mode hot we return error in case request wants data, that we've already rotated
 	if g.config.StoreMode == StoreModeHot {
-		if g.fracManager.Flags().IsCapacityExceeded() && g.earlierThanOldestFrac(uint64(from)) {
+		if g.fracManager.Flags().IsCapacityExceeded() && g.earlierThanOldestFrac(uint64(req.From)) {
 			metric.RejectedRequests.WithLabelValues("search", "old_data").Inc()
 			return &storeapi.SearchResponse{Code: storeapi.SearchErrorCode_INGESTOR_QUERY_WANTS_OLD_DATA}, nil
 		}


dkharms:
@coderabbitai review

eguguchkin:
nit:

> Suggestion: we should have a development guidlines file where we could store rules like "counters must end with _total" and "metrics should never end with _total"

+1 

Is .github/CONTRIBUTING.md the right place for this?

frac/processor/search_stats.go
```diff
@@ -17,32 +17,37 @@ var (
...
-		Name:      "agg_nodes_total",
+		Name:      "agg_nodes",
 		Buckets:   prometheus.ExponentialBuckets(1, 2, 20),
+		Help:      "Number of aggregation nodes in search query tree per search",
 	})
 	searchHitsTotal = promauto.NewHistogram(prometheus.HistogramOpts{
 		Namespace: metricsNamespace,
 		Subsystem: metricsSubsystem,
-		Name:      "hits_total",
+		Name:      "hits",
```
eguguchkin: The name "hist" refers to the metric type (histogram), not its connection to queries. This metric actually shows the distribution of the query result size, or more specifically, the number of IDs scanned per query.

frac/processor/search_stats.go
```diff
@@ -17,32 +17,37 @@ var (
...
-		Name:      "nodes_total",
+		Name:      "nodes",
 		Buckets:   prometheus.ExponentialBuckets(1, 2, 16),
+		Help:      "Number of nodes in search query tree per search",
 	})
 	searchSourcesTotal = promauto.NewHistogram(prometheus.HistogramOpts{
 		Namespace: metricsNamespace,
 		Subsystem: metricsSubsystem,
-		Name:      "sources_total",
+		Name:      "sources",
```
eguguchkin: Since we're renaming it anyway, let's pick a more descriptive name connected with lids.

frac/fraction.go
```diff
@@ -32,24 +32,28 @@ var (
 		Subsystem: "fetcher",
 		Name:      "fraction_stages_seconds",
 		Buckets:   metric.SecondsBuckets,
+		Help:      "Fetch processing time by stage",
 	}, []string{"stage", "fraction_type"})
 	fractionAggSearchSec = promauto.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: "seq_db_store",
 		Subsystem: "search",
-		Name:      "tracer_fraction_agg_search_sec",
+		Name:      "tracer_fraction_agg_search_seconds",
```
eguguchkin: Lets remove "tracer_" prefix? Here and below

cheb0:
Added a new section "General Coding Guidelines" to `CONTRIBUTING.md` with a point that metrics must follow the convention.

metric/metric.go
```diff
@@ -10,23 +10,23 @@ import (
 var (
 	Version = promauto.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "seq_db",
-		Name:      "version",
-		Help:      "",
+		Name:      "version_info_total",
```
eguguchkin: I think we might be using a counter here unnecessarily? Maybe we should have used a gauge instead?

Or rename the metric to version_starts_total.

metric/metric.go
```diff
@@ -10,23 +10,23 @@ import (
...
+		Help:      "Number of seq-db instances with a particular version",
```
eguguchkin: And accordingly, the description is something like: always 1 for the current version.

metric/store.go
```diff
@@ -10,126 +10,126 @@ var (
...
+		Help:      "Byte size of document in bulk requests",
 		Buckets:   prometheus.ExponentialBuckets(1, 4, 16),
 	})
 	BulkMetaBytesTotal = promauto.NewHistogram(prometheus.HistogramOpts{
 		Namespace: "seq_db_store",
 		Subsystem: "bulk",
-		Name:      "meta_bytes_total",
-		Help:      "",
+		Name:      "meta_bytes",
+		Help:      "Byte size of metadata in bulk requests in bytes",
```
eguguchkin: Rabbit is right about this.



image

proxy/bulk/metrics.go
```diff
@@ -9,8 +9,8 @@ var (
 	inflightBulks = promauto.NewGauge(prometheus.GaugeOpts{
 		Namespace: "seq_db_ingestor",
 		Subsystem: "bulk",
-		Name:      "in_flight_queries_total",
-		Help:      "",
+		Name:      "in_flight_queries",
```
eguguchkin: I agree with rabbit here as well. And probably we can call it just `in_flight`.

storeapi/grpc_bulk.go
```diff
@@ -20,7 +20,8 @@ import (
 var inflightBulksTotal = promauto.NewGauge(prometheus.GaugeOpts{
 	Namespace: "seq_db_store",
 	Subsystem: "bulk",
-	Name:      "in_flight_queries_total",
+	Name:      "in_flight_queries",
```
eguguchkin: we can just call it `in_flight` (since we're changing it anyway)

and code rabbit also suggested a fix in Help

frac/fraction.go
cheb0: done

frac/processor/search_stats.go
cheb0: from now on it's `lids_loaded`

metric/metric.go
cheb0: thing is, I'd avoid changing type since we need to maintain both old and new in a single dashboard. I'm not sure if we can do that. Maybe in a different PR

metric/metric.go
cheb0: updated a desc

dkharms:
@seqbench up main

dkharms:
Oh-oh, @dkharms `('>o.o)'>`!

Something went wrong and I couldn't process you request.
Please take a closer look at error message:

```
cannot apply manifest for baseline: docker-compose up failed: exit status 1
```

dkharms:
Nice, @dkharms `<(-^,^-)=b`!

Your request was successfully served.
Have a great time!

PR #291, opened 2025-12-03
feat: support nanosecond MID (phase 2)

### LINKED ISSUE #232: feat: support microsecond resolution for stored documents

cheb0:
# Description

Main differences with [the previous PR](https://github.com/ozontech/seq-db/pull/248):
- all `seq.MID` are now in nanoseconds.
- all new versions (frac version, meta version, etc) with nano MID are used everywhere by default.
- proxy is able to extract nanosecond time from JSON docs now.
- store protocol is now `2` which indicates that MID are sent in nanoseconds
- frac info cache stores info as json, `MID` are converted to milliseconds. `Info` now has custom JSON marshaling methods

Fixes #232

---

### DIFF
--- a/proxy/search/ingestor.go
+++ b/proxy/search/ingestor.go
@@ -513,7 +513,7 @@ func responseToQPR(resp *storeapi.SearchResponse, source uint64, explain bool) *
 			pbhist := bin.Hist
 
 			tbin := seq.AggBin{
-				MID:   seq.MillisToMID(uint64(bin.Ts.AsTime().UnixMilli())),
+				MID:   seq.MID(bin.Ts.AsTime().UnixNano()),
 				Token: bin.Label,
 			}
 
--- a/proxyapi/grpc_get_aggregation_test.go
+++ b/proxyapi/grpc_get_aggregation_test.go
@@ -92,8 +92,8 @@ func prepareGetAggregationTestData(t *testing.T, cData getAggregationTestCaseDat
 		sr := &search.SearchRequest{
 			Explain: req.Query.Explain,
 			Q:       []byte(req.Query.Query),
-			From:    seq.MillisToMID(uint64(req.Query.From.AsTime().UnixMilli())),
-			To:      seq.MillisToMID(uint64(req.Query.To.AsTime().UnixMilli())),
+			From:    seq.MID(req.Query.From.AsTime().UnixNano()),
+			To:      seq.MID(req.Query.To.AsTime().UnixNano()),
 		}
 		if len(cData.aggQ) > 0 {
 			for _, query := range cData.aggQ {
--- a/proxyapi/grpc_get_histogram_test.go
+++ b/proxyapi/grpc_get_histogram_test.go
@@ -98,8 +98,8 @@ func prepareGetHistogramTestData(t *testing.T, cData getHistogramTestCaseData) g
 		sr := &search.SearchRequest{
 			Explain:  req.Query.Explain,
 			Q:        []byte(req.Query.Query),
-			From:     seq.MillisToMID(uint64(req.Query.From.AsTime().UnixMilli())),
-			To:       seq.MillisToMID(uint64(req.Query.To.AsTime().UnixMilli())),
+			From:     seq.MID(req.Query.From.AsTime().UnixNano()),
+			To:       seq.MID(req.Query.To.AsTime().UnixNano()),
 			Interval: seq.DurationToMID(intervalDur),
 		}
 		siSearchMock = &siSearchMockData{
--- a/proxyapi/grpc_search_test.go
+++ b/proxyapi/grpc_search_test.go
@@ -88,8 +88,8 @@ func prepareSearchTestData(t *testing.T, cData searchTestCaseData) searchTestDat
 			Q:           []byte(req.Query.Query),
 			Size:        int(req.Size),
 			Offset:      int(req.Offset),
-			From:        seq.MillisToMID(uint64(req.Query.From.AsTime().UnixMilli())),
-			To:          seq.MillisToMID(uint64(req.Query.To.AsTime().UnixMilli())),
+			From:        seq.MID(req.Query.From.AsTime().UnixNano()),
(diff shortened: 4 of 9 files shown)


frac/common/info.go
```diff
@@ -117,3 +118,35 @@ func (s *Info) IsIntersecting(from, to seq.MID) bool {
 	// check with distribution
 	return s.Distribution.IsIntersecting(from, to)
 }
+
+// MarshalJSON implements custom JSON marshaling to always store From and To in milliseconds
+func (s *Info) MarshalJSON() ([]byte, error) {
+	type InfoAlias Info // type alias to avoid infinite recursion
```
eguguchkin: nit: Maybe `TmpInfo` for consistency?

frac/processor/search.go
```diff
@@ -206,7 +206,7 @@ func iterateEvalTree(
 						zap.Time("mid", mid.Time()))
 					continue
 				}
-				bucketIndex := uint64(mid)/params.HistInterval - histBase
+				bucketIndex := seq.MIDToMillis(mid)/params.HistInterval - histBase
```
eguguchkin: nit:

For getting bucketIndex, MID values can be converted to mills, or the interval can be converted to MID.

It seems that the first option is better - we won’t need to convert MID each time in a loop.

Also, consistency matters: Earlier, during conversion into the map, and later in the qpr merge method, we convert the interval to MID. If we do the same here, the code will be more consistent.

proxy/search/ingestor.go
```diff
@@ -513,7 +513,7 @@ func responseToQPR(resp *storeapi.SearchResponse, source uint64, explain bool) *
 			pbhist := bin.Hist
 
 			tbin := seq.AggBin{
-				MID:   seq.MID(bin.Ts.AsTime().UnixMilli()),
+				MID:   seq.MillisToMID(uint64(bin.Ts.AsTime().UnixMilli())),
```
eguguchkin: nit:

Here and in many other similar places, we could use something like `seq.MID(bin.Ts.AsTime().UnixNano())` instead of `seq.MillisToMID(uint64(bin.Ts.AsTime().UnixMilli()))`.

It's more explicit (in my opinion) and less redundant (avoids an extra call for conversion to MID).

But maybe you wanted a way to quickly find all the places where we added the conversion?

proxyapi/grpc_get_aggregation_test.go
```diff
@@ -92,8 +92,8 @@ func prepareGetAggregationTestData(t *testing.T, cData getAggregationTestCaseDat
 		sr := &search.SearchRequest{
 			Explain: req.Query.Explain,
 			Q:       []byte(req.Query.Query),
-			From:    seq.MID(req.Query.From.AsTime().UnixMilli()),
-			To:      seq.MID(req.Query.To.AsTime().UnixMilli()),
+			From:    seq.MillisToMID(uint64(req.Query.From.AsTime().UnixMilli())),
```
eguguchkin: nit: seq.MID(req.Query.From.AsTime().UnixNano()),

proxyapi/grpc_get_histogram_test.go
```diff
@@ -98,8 +98,8 @@ func prepareGetHistogramTestData(t *testing.T, cData getHistogramTestCaseData) g
 		sr := &search.SearchRequest{
 			Explain:  req.Query.Explain,
 			Q:        []byte(req.Query.Query),
-			From:     seq.MID(req.Query.From.AsTime().UnixMilli()),
-			To:       seq.MID(req.Query.To.AsTime().UnixMilli()),
+			From:     seq.MillisToMID(uint64(req.Query.From.AsTime().UnixMilli())),
```
eguguchkin: nit: seq.MID(req.Query.From.AsTime().UnixNano()),

proxyapi/grpc_search_test.go
```diff
@@ -88,8 +88,8 @@ func prepareSearchTestData(t *testing.T, cData searchTestCaseData) searchTestDat
 			Q:           []byte(req.Query.Query),
 			Size:        int(req.Size),
 			Offset:      int(req.Offset),
-			From:        seq.MID(req.Query.From.AsTime().UnixMilli()),
-			To:          seq.MID(req.Query.To.AsTime().UnixMilli()),
+			From:        seq.MillisToMID(uint64(req.Query.From.AsTime().UnixMilli())),
```
eguguchkin: nit: seq.MID(req.Query.From.AsTime().UnixNano()),

frac/common/info.go
cheb0: fixed

frac/processor/search.go

proxy/search/ingestor.go

there was no particular reason other than just "straightforwardness" of wrapping `bin.Ts.AsTime().UnixMilli()` in `seq.MillisToMID`
I agree it's more clean to just have `seq.MID(bin.Ts.AsTime().UnixNano())` and I don't see any potential pitfalls

forshev (approved):

proxyapi/grpc_v1.go
```diff
@@ -329,7 +329,7 @@ func convertAggsQuery(aggs []*seqproxyapi.AggQuery) ([]search.AggQuery, error) {
 			)
 		}
 
-		aggQuery.Interval = seq.MID(interval.Milliseconds())
+		aggQuery.Interval = seq.MID(interval.Nanoseconds())
```
eguguchkin: получается что все таки от прокси к стору могут приходить значения разной размерности

в фазе 1 милисекунды

в фазе 2 наносекунды

но мы не проверяем на сторе какой версии запрос пришел

eguguchkin (approved):

PR #298, opened 2025-12-19
feat: uniqCount aggregation

### LINKED ISSUE #295: Add UniqCount aggregation function
Currently, we have `Count` function. It was the only function we had at the beginning in `seq-db`. Currently, it  works as follows:
`select count(foo) group by foo`
which means we can only group and count over the same field.

Another aggregation function we have is `Unique`. It works like `Count` except it doesn't count unique values but only finds them.
 
New aggregation function `UniqCount` should allow to group by some field and aggregate over another field. It will work as follows:
`select UniqCount(foo) group by bar`

cheb0:
# Description

Fixes #295

---

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
       - name: Deploy latest (unstable) version of seq-db
         run: |
-          curl --insecure -X POST "https://${{ secrets.CONTINUOUS_BENCHMARKS_ENDPOINT }}/seqagent/v1/b/up" \
+          curl -fX POST "https://${{ secrets.CONTINUOUS_BENCHMARKS_ENDPOINT }}/seqagent/v1/b/up" \
             -H "Content-Type: application/json" \
             -H "X-Seqbench-Secret: ${{ secrets.SEQBENCH_SECRET }}" \
             -d @- <<EOF


proxy/search/ingestor.go
```diff
@@ -582,6 +594,9 @@ func (si *Ingestor) searchShard(
 			if errMessage == consts.ErrTooManyFieldTokens.Error() {
 				return nil, source, fmt.Errorf("store forbids aggregation request: %w", consts.ErrTooManyFieldTokens)
 			}
+			if errMessage == consts.ErrTooManyFieldValues.Error() {
+				return nil, source, fmt.Errorf("store forbids aggregation request: %w", consts.ErrTooManyFieldValues)
+			}
```
forshev, line 597: we need to check new resp.Code later as well. on line 614-625

proxy/search/ingestor.go
```diff
@@ -517,19 +519,29 @@ func responseToQPR(resp *storeapi.SearchResponse, source uint64, explain bool) *
...
 				Samples:   pbhist.Samples,
 				NotExists: pbhist.NotExists,
+				Values:    values,
 			}
 		}
 
 		aggs[i] = seq.AggregatableSamples{
 			SamplesByBin: to,
 			NotExists:    agg.NotExists,
+			ValuesPool:   stringPool,
```
forshev: nit: seem like we can just do `ValuesPool:   agg.ValuesPool`, and remove `stringPool` variable

seq/qpr.go
```diff
@@ -322,6 +363,14 @@ func (h *SamplesContainer) Merge(hist *SamplesContainer) {
 	for _, v := range hist.Samples {
 		h.InsertSample(v)
 	}
+	if len(hist.Values) > 0 {
+		if h.Values == nil {
+			h.Values = make(map[uint32]struct{}, len(hist.Values))
+		}
+		for idx := range hist.Values {
+			h.Values[idx] = struct{}{}
+		}
```
forshev: nit: seems like we can use `maps.Copy` here

seq/qpr.go
```diff
@@ -241,11 +245,46 @@ func (q *AggregatableSamples) Merge(agg AggregatableSamples) {
 		q.SamplesByBin = make(map[AggBin]*SamplesContainer, len(agg.SamplesByBin))
 	}
 
+	oldToNewIndexMap := make(map[uint32]uint32, len(agg.ValuesPool))
```
eguguchkin: nit: Could we use a slice instead of a map here? (It would likely perform better)

proxy/search/ingestor.go
cheb0: fixed. also added a new error to proto store API

proxy/search/ingestor.go

seq/qpr.go
cheb0: fixed

PR #305, opened 2025-12-30
chore: track frac search wait time in explain

cheb0:
# Description

For some search requests, trace file looks like that:
Screenshot_20251230_113216

i.e. there are holes when we wait for wait group in Searcher. I decided to expand explain entry as follows:

```
          {
            "duration": "302.14541ms",
            "message": "store/Search",
            "children": [
              {
                "duration": "17.553µs",
                "message": "parse query"
              },
              {
                "duration": "301.703205ms",
                "message": "search iteratively"
              },
              {
                "duration": "928.536095ms",
                "message": "search iteratively (cpu time)"
              },
              {
                "duration": "265.08246ms",
                "message": "waiting goroutines (all cores)"
              }
            ]
          }
```

i.e. number `waiting goroutines` shows how much time we spent waiting and we _could_ do (and probable did) other work. However, if we processing a single search request at a time, then this time is essentially wasted. Obviously, this metric is not absolutely honest, system might have been significantly overloaded.

Numbers track times across all cores and do look odd, but I didn't find any other better way. I suppose we could do that, because explain entries are not loaded to tracing storage and we do not visualize them anywhere.

---


fracmanager/searcher.go
```diff
@@ -120,6 +142,11 @@ func (s *Searcher) searchDocsAsync(ctx context.Context, fracs []frac.Fraction, p
 	wg := sync.WaitGroup{}
 	qprs := make([]*seq.QPR, len(fracs))
 
+	var totalSearchTimeNanos atomic.Int64
+	var totalWaitTimeNanos atomic.Int64
```
eguguchkin: nit: It seems that totalWaitTimeNanos does not necessarily need to be an atomic type.

fracmanager/searcher.go
```diff
@@ -129,26 +156,39 @@ loop:
...
+				searchStart := time.Now()
 				var fracErr error
 				if qprs[i], fracErr = s.fracSearch(ctx, params, frac); fracErr != nil {
 					once.Do(func() {
 						err = fracErr
 						cancel()
 					})
 				}
+				searchDuration := time.Since(searchStart)
+				totalSearchTimeNanos.Add(searchDuration.Nanoseconds())
```
eguguchkin: nit: We could avoid using atomics altogether by storing start times in a slice, just like we do with end times, and then compute the difference (similar to how you calculate totalWaitTimeNanos).

fracmanager/searcher.go
```diff
@@ -129,26 +156,39 @@ loop:
...
 				wg.Done()
 			}()
 		}
 	}
 
 	wg.Wait()
+	waitEndTime := time.Now()
+
+	for _, doneTime := range wgDoneTimes {
+		waitDuration := waitEndTime.Sub(doneTime)
```
eguguchkin: nit: theoretically, we might have 0 in wgDoneTimes if fracSearch returns an error.

fracmanager/searcher.go
```diff
@@ -69,6 +78,19 @@ func (s *Searcher) SearchDocs(ctx context.Context, fracs []frac.Fraction, params
 		subSearchesCnt++
 	}
 
+	if tr != nil && tr.Enabled() {
```
eguguchkin: nit: `tr != nil` is a redundant check, see the Enabled method.

fracmanager/searcher.go
cheb0: done

fracmanager/searcher.go
cheb0: checking for 0 now

PR #328, opened 2026-01-26
perf: implement bitpack encoding for LID and MID blocks

### LINKED ISSUE #312: Use bitpack encoding for LID blocks
LIDs (posting lists) blocks are the most crucial data structure for search along with tokens. Currently, we utilize varints, which are slow and inefficient. 

The proposed solution is to use bitpack encoding. There are two libraries to test in Go:
- ready to use library [intcomp](https://github.com/ronanh/intcomp) which works on top of 128 numbers chunks. Bit width is defined for every 32 numbers.
- Daniel Lemire's fastPFOR delta which is [ported to Go](https://github.com/zentures/encoding/blob/master/fastpfor/fastpfor.go). It tries to find optimal bit width for 128 numbers at the same time cutting large detas and placing them in a footer of the page. The algorithm is tuned for 64k pages

Need to test: 
- dataset size overhead
- search and throughput performance metrics
- estimate how much is the overhead if we disable zstd for LID blocks

cheb0:
# Description

Replaces varint encoding with faster delta bitpacking. Both LID and MID blocks now use bitpack. Currently, [intcomp](https://github.com/ronanh/intcomp) library is used. The lib doesn't utilize SIMD, so we might update to something else in future.

### Measurements

**compression**
bitpack compresses a lot better: for varints zstd compresses with ratio **~1.7-2.0** while it only compresses delta bitpacked data with ratio **~1.3**. Therefore we potentially can disable zstd on benchmarks with a slight dataset size overhead.

**dataset size**
Overall, we reach approximately same dataset size. For some envs there is a small benefit of around **-3%** of total dataset.

**Search latency (stg fractions repacked with bitpack)**

<!DOCTYPE html>

<html>
<body>

Query | Type | Ids | Total | cold, ms | hot, ms | cold (branch), ms | hot (branch), ms | cold diff | hot diff
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
service:W AND message:"A B C D" (found) | reg | 100 | 0 | 5.68 | 0.13 | 5.29 | 0.08 | **-6.9%** | -38.5%
service:W AND message:"A B C D" (not found) | reg | 0 | 0 | 37.27 | 0.53 | 16.19 | 0.55 | **-56.6%** | 3.8%
service:X AND level:3 AND NOT (x6) | reg | 0 | 0 | 66.21 | 2.13 | 32.39 | 2.02 | **-51.1%** | -5.2%
service:X AND level:[0 to 4] | hist | 100 | 1617439 | 185.71 | 50.86 | 99.57 | 42.61 | **-46.4%** | -16.2%
zone:Z AND service:X AND level:4 AND job_id:J AND message:"X Y Z" | hist | 100 | 1791 | 76.59 | 5.13 | 32.4 | 4.81 | **-57.7%** | -6.2%
k8s_container:C AND k8s_namespace:N AND service:X AND level:4 AND source:S | hist | 100 | 11798 | 93.15 | 18.97 | 48.08 | 18 | **-48.4%** | -5.1%
service:W AND message:"X Y Z A" AND logger:L AND source:S | hist | 100 | 260 | 68.24 | 2.52 | 34.45 | 1.51 | **-49.5%** | -40.1%
service:W \| group by k8s_pod | agg | 100 | 170764 | 259.34 | 188.19 | 231.81 | 190 | **-10.6%** | 1%
service:O \| group by k8s_pod | agg | 100 | 1619624 | 273.8 | 199.66 | 235.28 | 201.4 | **-14.1%** | 0.9%
service:W \| group by k8s_pod (timeseries) | agg | 100 | 170764 | 467.04 | 274.9 | 405.11 | 268.39 | **-13.3%** | -2.4%

</body>

</html>

Fixes #312 

---


cheb0:
@seqbenchbot up main search-logbench

cheb0:
@seqbenchbot up main bulk

cheb0:
@seqbenchbot down dc2d4d40

cheb0:
@seqbenchbot up main search-logbench

util/algorithms.go
```diff
@@ -0,0 +1,110 @@
...
+	for i := 0; i < n; i++ {
+		res[i] = binary.LittleEndian.Uint32(buf)
+		buf = buf[sizeOfUint32:]
+	}
+	return res
+}
+
+// CastAsUint64 allows working on []byte slice as []uint64. Uses unsafe casts for little endian, allocates
+// a new buf and copies on big endian hosts. The caller must work as read only.
+func CastAsUint64(buf []byte) []uint64 {
```
dkharms: I guess it is worth putting a check that `(n % sizeOfUint64) == 0`. Just in case.

util/algorithms.go
```diff
@@ -0,0 +1,110 @@
...
+	n := len(buf) / sizeOfUint64
+	if n == 0 {
+		return nil
+	}
+
+	if littleEndian {
+		return unsafe.Slice((*uint64)(unsafe.Pointer(unsafe.SliceData(buf))), n)
```
dkharms: Are there any potential issues with alignment? AFAIK, there is no alignment guarantees that are provided by the Golang memory allocator.

For x86-64 it is not a problem (well, performance-wise it is), but for other architectures it is.

Maybe, we should add another check, something like:
```
assert(uintptr(unsafe.Pointer(&buf[0])) % unsafe.Alignof(uint64(0)) == 0)
```

frac/sealed/lids/block.go
```diff
@@ -49,13 +44,46 @@ func (b *Block) GetSizeBytes() int {
...
+	if fracVer >= config.BinaryDataV3 {
+		return b.unpackBitpack(data, buf)
+	}
+
+	return b.unpackVarint(data, buf)
+}
+
+func (b *Block) unpackBitpack(data []byte, buf *UnpackBuffer) error {
+	if data[0] == 1 {
+		b.IsLastLID = true
```
dkharms: IIRC `IsLastLID` only affects packing and unpacking -- specifically, when we build slice of offsets for chunks of lids. This field in not accessed within search queries at all.

So maybe we can omit it? You store offsets as a part if LID block now so basically this field is useless.

util/algorithms.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+func detectLittleEndian() bool {
+	var i int32 = 1
+	b := *(*byte)(unsafe.Pointer(&i))
+	return b == 1
+}
+
+// CopyUints32 copies srt to dst byte slice. If host is little-endian, then uses direct memory copy instead of loop.
+func CopyUints32(src []uint32, dst []byte) []byte {
```
dkharms: Have you measured impact of reinterpret cast for such functions?

forshev (approved):

util/algorithms.go
cheb0: I deleted the entire file. I think it's too early for those optimizations. Maybe will do in a separate PR.

frac/sealed/lids/block.go
cheb0: It's used by index analyzer cmd file. Probably deletion a separate PR would be fine, now it's just easier for me to avoid thinking about those fields.

cheb0: Deleted it from block format. We can rely on `MinTID/MaxTID` in `index_analyzer`, so we can delete this field altogether after index split is merged.

dkharms (approved):

cheb0:
@seqbenchbot up main bulk

cheb0:
@seqbenchbot down d9e8e68e

PR #334, opened 2026-01-30
feat: new WAL file for meta

### LINKED ISSUE #311: Make WAL resilient to hardware crash
Currently, docs and meta file writer (essentially WAL), can't handle hardware crash. For example, server may shut down with power off switch.  Since, we can't write large doc blocks atomically, such blocks might be partially written even in presence of calling `fsync` system call.

If this happens, WAL structure might not be complete, and store panics on active fraction replay. The proposed solution is to redesign WAL so that it's able to cope with instant power shutdown.

cheb0:
# Description

A new WAL file which can withstand arbitrary failures like power shutdowns, lost disk sector writes or corrupted bytes.

There is also a new `MetaBlock` type which replaces `DocBlock` at all places where meta is transfered or stored to filesystem.

Fixes #311 

---

### DIFF
--- a/storage/wal_reader.go
+++ b/storage/wal_reader.go
@@ -85,8 +85,11 @@ func (r *WalReader) Entries() iter.Seq[WalRecord] {
 			}
 
 			if errors.Is(err, io.EOF) || n < WalBlockHeaderLen {
-				startCorruptionTracking(offset)
-				logCorruptionEnd(offset + int64(n))
+				// log corruption only if at least 1 was read
+				if n > 0 {
+					startCorruptionTracking(offset)
+					logCorruptionEnd(offset + int64(n))
+				}
 				return
 			}
 


cheb0:
@seqbenchbot up main bulk

cheb0:
@seqbenchbot down 0c97854c

storage/wal_reader.go
```diff
@@ -0,0 +1,154 @@
...
+func NewWalReader(limiter *ReadLimiter, reader io.ReaderAt, baseFileName string) (*WalReader, error) {
+	header := make([]byte, WALHeaderSize)
+	n, err := limiter.ReadAt(reader, header, 0)
+	if err != nil && !errors.Is(err, io.EOF) {
+		return nil, fmt.Errorf("failed to read WAL header: %w", err)
+	}
+	if n < WALHeaderSize {
+		return nil, fmt.Errorf("WAL file too short: expected at least %d bytes, got %d", WALHeaderSize, n)
+	}
+	magic := binary.LittleEndian.Uint32(header[0:4])
```
dkharms: nit:
```suggestion
	magic := binary.LittleEndian.Uint32(header[0:unsafe.Sizeof(WALMagic)])
```

storage/meta_block.go
```diff
@@ -0,0 +1,215 @@
...
+	offsetMetaBlockHeaderChecksum  = 23 // 4 bytes (H) Header checksum - covers bytes 0-22
+
+	MetaBlockHeaderLen      = 27
+	MetaBlockCurrentVersion = uint8(1)
+)
+
+// MetaBlock format: M : V : C : LLLL : UUUU : PPPP : DDDD-DDDD : HHHH
+// M = Magic (101), V = Version, C = Codec, L = Length, U = Raw Length, P = Payload Checksum, D = Docs Offset, H = Header Checksum
+
+type MetaBlock []byte
```
dkharms: Maybe we can go with `WALBlock`?

storage/wal_writer.go
```diff
@@ -0,0 +1,165 @@
...
+	WALVersion1 uint8 = 1
+	// WALCurrentVersion is the current WAL format version.
+	WALCurrentVersion = WALVersion1
+	// WALHeaderSize is the size of the WAL header in bytes (4 bytes magic + 1 byte version). 59 bytes are also reserved
+	// due to alignment
+	WALHeaderSize = 5
+	// BlockAlignment is the alignment boundary for blocks in the new WAL format. Must be greater than
+	// MetaBlock header (27 bytes) to prevent header torn writes and allow faster navigation during replay
+	// of corrupted WAL file
+	BlockAlignment int64 = 64
```
dkharms:
```suggestion
	WALBlockAlignment int64 = 64
```

storage/wal_reader.go
```diff
@@ -0,0 +1,154 @@
...
+	"fmt"
+	"io"
+	"iter"
+
+	"go.uber.org/zap"
+
+	"github.com/ozontech/seq-db/logger"
+)
+
+type WalRecord struct {
```
dkharms, line 16: Let's name everything WAL-related consistently. Somewhere you have `Wal`, somewhere you have `WAL`.

storage/wal_reader.go
```diff
@@ -0,0 +1,154 @@
...
+			}
+		}
+		startCorruptionTracking := func(offset int64) {
+			if corruptionStart < 0 {
+				corruptionStart = offset
+			}
+		}
+
+		for {
+			headerBuf := make([]byte, MetaBlockHeaderLen)
```
dkharms: Seems like we could reuse `headerBuf` on each iteration.
Is it possible to move `headerBuf` outside of the loop?

storage/wal_writer.go
```diff
@@ -0,0 +1,165 @@
...
+	"sync"
+	"sync/atomic"
+
+	"go.uber.org/zap"
+
+	"github.com/ozontech/seq-db/logger"
+	"github.com/ozontech/seq-db/metric/stopwatch"
+)
+
+const (
```
dkharms, line 19: Maybe we can move these variables to `meta_block.go` (which must be renamed to `wal_block.go` in my opinion) and introduce new structure `WALHeader` with different access methods e.g. `Magic()` and `Version()`?

storage/wal_writer.go
```diff
@@ -0,0 +1,165 @@
...
+	// WALMagic is the magic number at the start of WAL files
+	WALMagic uint32 = 0xFFFFFFFF
+	// WALVersion1 is the first version of WAL file with CRC32 checksums and 64 byte alignment for blocks.
+	WALVersion1 uint8 = 1
```
dkharms: Why `WALVersion1` and not just `WALVersion`?

storeapi/grpc_bulk.go
```diff
@@ -64,6 +65,11 @@ func (g *GrpcV1) doBulk(ctx context.Context, req *storeapi.BulkRequest) error {
 
 	start := time.Now()
 
+	if !storage.IsMetaBlock(req.Metas) {
+		// We use MetaBlock now for metas everywhere, DocBlock might be sent by legacy proxy service.
+		req.Metas = storage.PackDocBlockToMetaBlock(req.Metas)
```
dkharms: Why did you decide to calculate checksums on store and not on proxy?
This seems less efficient (we have to calculate the same values replica-factor times).

Yeah, this is part of code is still necessary since there is a period when new version of store and old version of proxy are deployed. But when new version of proxy is deployed we can move all that CPU-intensive calculations to proxy.

storage/wal_writer.go
```diff
@@ -0,0 +1,165 @@
...
+		curr := w.offset.Load()
+		nextSlotOffset := nextBlockOffset(curr)
+
+		if w.offset.CompareAndSwap(curr, nextSlotOffset+blockSize) {
+			result = nextSlotOffset
+			break
+		}
+	}
+	return result
+}
```
dkharms, line 62: I guess we can get rid of CAS loop where.

Let's rewrite initialization like this:
```go
func NewWalWriter(ws WriteSyncer, offset int64, skipSync bool) *WalWriter {
	w := &WalWriter{
		ws:       ws,
		skipSync: skipSync,
		notify:   make(chan struct{}, 1),
	}

 	// Here we have offset that can be calculated like 64 * `k` for some `k`.
	w.offset.Store(nextBlockOffset(offset))
	// write a header at the beggining if it's a new file
	if offset == 0 {
		if err := writeWALHeader(ws); err != nil {
			logger.Panic("failed to write WAL header", zap.Error(err))
		}

		if !skipSync {
			_ = ws.Sync()
		}


 	    // Here we have offset that can be calculated like 64 * `k` for some `k`.
        // In this case `k` is equal to 1.
		w.offset.Store(nextBlockOffset(WALHeaderSize)) 
	}

	w.wg.Add(1)
	go func() {
		w.syncLoop()
		w.wg.Done()
	}()

	return w
}
```

So we know that offset is already aligned.
Now we can reserve space like this:
```go
func (w *WalWriter) reserveSpace(blockSize int64) int64 {
	aligned := nextBlockOffset(blockSize)
    
    // w.offset is already aligned.
    // So when we add aligned block we still have aligned offset.
	end := w.offset.Add(aligned)
	start := end - aligned

	return start
}
```

I've applied these changes locally and tests are green.

frac/active.go
```diff
@@ -79,7 +81,34 @@ func NewActive(
...
+		walFileName := baseFileName + consts.WalFileSuffix
+		metaFile, metaStats = mustOpenFile(walFileName, config.SkipFsync)
+		metaSize = uint64(metaStats.Size())
+		writer = NewActiveWriter(docsFile, metaFile, docsStats.Size(), metaStats.Size(), config.SkipFsync)
+		var err error
+		walReader, err = storage.NewWalReader(readLimiter, metaFile, baseFileName)
+		if err != nil {
+			logger.Fatal("failed to initialize WAL reader", zap.String("fraction", baseFileName), zap.Error(err))
+		}
+	}
```
dkharms: Maybe we can move this to separate method?

storage/wal_writer.go
```diff
@@ -0,0 +1,165 @@
...
+type WriteSyncer interface {
+	io.ReaderAt
+	io.WriterAt
+	Sync() error
+}
+
+// WalWriter writes MetaBlocks to a WAL file with header and 64-byte alignment.
+// Format: [Header 5B] [... -> align to 64] [MetaBlock] [... -> align to 64] [MetaBlock] ...
+type WalWriter struct {
+	ws       WriteSyncer
```
dkharms: Why not reusing `frac.FileWriter` here? Yeah, I've noticed that in `FileWriter` there is a conversion from new `WAL` format to old `Meta` format but I guess it can be done somewhere else...

I do not like the duplication of fsync-coalescing logic here.

storage/meta_block.go
cheb0: renamed to `WalBlock`

storage/wal_writer.go
cheb0: fixed

storage/wal_reader.go
cheb0: done

storage/wal_reader.go
cheb0: done

storage/wal_writer.go
cheb0: The initial version had exactly that - a dedicated `WalHeader` struct. But I didn't see much of the point right now since there is no much usage of it.

I wouldn't move constants to `wal_block.go` since they are not related to the block itself, only to WAL.

storage/wal_writer.go
cheb0: Next line, `WALCurrentVersion = WalVersion1`

storage/wal_writer.go
cheb0: yep, if offset is always aligned, then CAS is not needed. fixed

storage/wal_writer.go
cheb0: That's a fair point. I redesigned a bit - `FileWriter` now (probably should be named `SyncFileWriter`) is not aware of any block formats, while convertions of doc => wal happens in a separate wrapper `LegacyMetaWriter`.

`WalWriter` now also works on top of `FileWriter`, which means it now reserves space from `FileWriter` too, since `offset` is owned by `FileWriter`.

Both `WalWriter` and `LegacyMetaWriter` implement a trait `MetaWriter`.

I also moved `FileWriter` to `storage` package to prevent circular dependency.

frac/active.go
cheb0: done

storage/wal_writer.go
dkharms: Yeah, I see, then a small nit:
```go
const (
	// _WalVersionInitial is the first version of WAL file
	// with CRC32 checksums and 64 byte alignment for blocks.
	_WalVersionInitial byte = iota
)

const (
	// WalMagic is the magic number at the start of WAL files.
	WalMagic uint32 = 0xFFFFFFFF
	// WalCurrentVersion is the current WAL format version.
	WalCurrentVersion = _WalVersionInitial
	// WalHeaderSize is the size of the WAL header in bytes (4 bytes magic + 1 byte version).
	// 59 bytes are also reserved due to alignment.
	WalHeaderSize = 5
)

const (
	// WalBlockAlignment is the alignment boundary for blocks in the new WAL format.
	// Must be greater than [WalBlockHeaderLen] to prevent header torn writes
	// and allow faster navigation during replay of corrupted WAL file.
	WalBlockAlignment int64 = 64
)
```

That's just my personal preference so you can ignore it (and it's very insignificant, I guess). I've added prefix (`_`) to `_WalVersionInitial` to send a signal to other developers that this variable should not be used.


By the way, you use specific WAL version instead of current WAL version in `wal_reader.go`:
```go
version := header[4]
if version != _WalVersionInitial {
	return nil, fmt.Errorf("unknown WAL version: %d (supported: %d)", version, _WalVersionInitial)
}
```

storage/wal_writer.go
dkharms: > Both `WalWriter` and `LegacyMetaWriter` implement a trait

I guess, you've confused languages a little bit :rofl: 
Overall, looks great.

cheb0:
@seqbenchbot up main bulk

cheb0:
@seqbenchbot down b9ff764f

frac/active.go
```diff
@@ -114,6 +118,35 @@ func NewActive(
...
+
+	if _, err := os.Stat(legacyMetaFileName); err == nil {
+		// .meta file exists
+		metaFile, metaStats := mustOpenFile(legacyMetaFileName, config.SkipFsync)
+		metaSize := uint64(metaStats.Size())
+		metaReader := storage.NewDocBlocksReader(readLimiter, metaFile)
+		writer := NewActiveWriterLegacy(docsFile, metaFile, docsStats.Size(), metaStats.Size(), config.SkipFsync)
+		logger.Info("using legacy meta file format", zap.String("fraction", baseFileName))
+		return metaFile, writer, &metaReader, nil, metaSize
+	} else {
```
eguguchkin: nit: else after return is not necessary

frac/active.go
```diff
@@ -133,6 +166,81 @@ func mustOpenFile(name string, skipFsync bool) (*os.File, os.FileInfo) {
 }
 
 func (f *Active) Replay(ctx context.Context) error {
```
eguguchkin, line 167: nit: we already check file existence during initialization in mustOpenMetaWriter, so we can simplify the check in Replay


```go
func (f *Active) Replay(ctx context.Context) error {
	if f.walReader != nil {
		return f.replayWalFile(ctx)
	}

	if f.metaReader != nil {
		return f.replayMetaFileLegacy(ctx)
	}

	logger.Info("neither wal nor legacy meta file was found, skipping replay", zap.String("fraction", f.BaseFileName))
	return nil
}
```

Since we always open either a meta file or a wal file during initialization, Replay could be even simpler:


```go
func (f *Active) Replay(ctx context.Context) error {
	if f.metaReader != nil {
		return f.replayMetaFileLegacy(ctx)
	}
	return f.replayWalFile(ctx)
}
```

frac/active.go
```diff
@@ -133,6 +166,81 @@ func mustOpenFile(name string, skipFsync bool) (*os.File, os.FileInfo) {
...
+	if _, err := os.Stat(metaFileName); err == nil {
+		return f.replayMetaFileLegacy(ctx)
+	}
+
+	logger.Info("neither wal nor legacy meta file was found, skipping replay", zap.String("fraction", f.BaseFileName))
+	return nil
+}
+
+func (f *Active) replayWalFile(ctx context.Context) error {
+	if f.walReader == nil {
```
eguguchkin: nit: this is a private method and its calls are under our control, so we don't need this check - we expect walReader to be initialized)

frac/active.go
```diff
@@ -133,6 +166,81 @@ func mustOpenFile(name string, skipFsync bool) (*os.File, os.FileInfo) {
...
 func (f *Active) Replay(ctx context.Context) error {
+	walFileName := f.BaseFileName + consts.WalFileSuffix
+	if _, err := os.Stat(walFileName); err == nil {
+		return f.replayWalFile(ctx)
+	}
+
+	metaFileName := f.BaseFileName + consts.MetaFileSuffix
+	if _, err := os.Stat(metaFileName); err == nil {
+		return f.replayMetaFileLegacy(ctx)
+	}
```
moflotas: nit: If crash happens and both `.wal` and `.meta` files are present, then `walReader` will not be initialized and store will fall with error

If we change order, first read `.meta`, then `.wal`, then it's fine

frac/file_writer_test.go
```diff
@@ -262,3 +264,40 @@ func TestSparseWrite(t *testing.T) {
 	e = os.Remove(rf.Name())
 	assert.NoError(t, e)
 }
+
+func TestLegacyMetaWriterConvertsWalBlockToDocBlock(t *testing.T) {
```
eguguchkin: all remaining file except this function duplicates storage/file_writer_test.go

storage/wal_writer.go
```diff
@@ -0,0 +1,80 @@
...
+	// due to alignment
+	WalHeaderSize = 5
+	// WalBlockAlignment is the alignment boundary for blocks in the new WAL format. Must be greater than
+	// WalBlock header (27 bytes) to prevent header torn writes and allow faster navigation during replay
+	// of corrupted WAL file
+	WalBlockAlignment int64 = 64
+)
+
+type WriteSyncer interface {
+	io.ReaderAt
```
eguguchkin: We don't actually use io.ReaderAt here. If we remove it, we'll end up with two identical interfaces in this package: WriteSyncer and fileWriterSyncer.

And we can remove one of them

storage/wal_reader.go
```diff
@@ -0,0 +1,155 @@
...
+		limiter:      limiter,
+		reader:       reader,
+		headerOffset: WalHeaderSize,
+		baseFileName: baseFileName,
+	}, nil
+}
+
+// Iter iterates through WAL file. Corrupted entries are skipped and never propagated to a client.
+// Corruption ranges are logged with "from" and "to" offsets.
+func (r *WalReader) Iter() iter.Seq[WalRecord] {
```
eguguchkin: nit: (naming) What if we call this method Blocks() or Entries()?

```
    for entry := range f.walReader.Entries() {
        // ... 
    }
```

Just a suggestion, feel free to ignore.

eguguchkin: nit: Another suggestion regarding error handling. When working with iterators, I use a pattern where the iterator provider also has a LastError() method:

```diff


-       for entry := range f.walReader.Iter() {
+       for entry := range f.walReader.Entries() {
                // Check for context cancellation
                select {
                case <-ctx.Done():
@@ -203,17 +203,13 @@ func (f *Active) replayWalFile(ctx context.Context) error {
                default:
                }

-               if entry.Err != nil {
-                       return entry.Err
-               }
-
                if uint64(entry.Offset) > next {
                        next += step
                        progress := float64(uint64(entry.Offset)) / float64(f.info.MetaOnDisk) * 100
                        logger.Info("replaying batch, meta",
                                zap.String("name", f.info.Name()),
                                zap.Int64("from", entry.Offset),
-                               zap.Int64("to", entry.Offset+entry.Size),
+                               zap.Int64("to", entry.Offset+int64(entry.Data.FullLen())),
                                zap.Uint64("target", f.info.MetaOnDisk),
                                util.ZapFloat64WithPrec("progress_percentage", progress, 2),
                        )
@@ -223,6 +219,10 @@ func (f *Active) replayWalFile(ctx context.Context) error {
                f.indexer.Index(f, entry.Data, &wg, sw)
        }

+       if err := f.walReader.LastError(); err != nil {
+               return err
+       }
+
        wg.Wait()
```

This way, errors can be checked after the iteration loop instead of handling them inside each iteration.


```diff
        Offset int64
-       Size   int64
-       Err    error
 }
```

```diff

                        if err != nil && !errors.Is(err, io.EOF) {
                                logCorruptionEnd(offset)
-                               yield(WalRecord{Offset: offset, Err: err})
+                               r.lastError := err
                                return
                        }
```

Feel free to ignore.

storage/wal_reader.go
```diff
@@ -0,0 +1,155 @@
...
+
+			// header is correct, try to read the payload
+			blockLen := int64(mb.FullLen())
+			blockBuf := make([]byte, blockLen)
+			n, err = r.limiter.ReadAt(r.reader, blockBuf, offset)
+
+			if err != nil && !errors.Is(err, io.EOF) {
+				// this is the last WAL record
+				// start corruption tracking if not started already and print
+				startCorruptionTracking(offset)
```
eguguchkin: I would distinguish between two cases:

1. When we read but failed to parse (this indicates file corruption).
2. When we couldn't read for some other reason (hardware failure?).
 

In the first case, the file is corrupted, and it makes sense to track this in logs via `startCorruptionTracking()`.

In the second case, we simply propagate the error upwards (as done here), and most likely the service will be shut down further up the stack.

And it seems the second case? (so `startCorruptionTracking()` no need here)

So the comment `// this is the last WAL record` doesn't match reality here. This is more relevant to the next check: `if errors.Is(err, io.EOF) || int64(n) < blockLen {`

Also, it's unclear what we expect to see in the "from" and "to" fields in the logs. 
That's why calling `startCorruptionTracking()` is unnecessary here.

storage/wal_reader.go
```diff
@@ -0,0 +1,155 @@
...
+				logCorruptionEnd(offset)
+				yield(WalRecord{Offset: offset, Err: err})
+				return
+			}
+
+			if errors.Is(err, io.EOF) || int64(n) < blockLen {
+				startCorruptionTracking(offset)
+				logCorruptionEnd(offset)
```
eguguchkin: The same question: what we expect to see in the "from" and "to" field here

Maybe 

```go
				logCorruptionEnd(offset + n)
```

storage/wal_reader.go
```diff
@@ -0,0 +1,155 @@
...
+			n, err := r.limiter.ReadAt(r.reader, headerBuf, offset)
+
+			if err != nil && !errors.Is(err, io.EOF) {
+				logCorruptionEnd(offset)
+				yield(WalRecord{Offset: offset, Err: err})
+				return
+			}
+
+			if errors.Is(err, io.EOF) || n < WalBlockHeaderLen {
+				logCorruptionEnd(offset)
```
eguguchkin: And it looks like we need call 				

``` go
    startCorruptionTracking(offset)
    logCorruptionEnd(offset + n)
```

storage/wal_reader.go
```diff
@@ -0,0 +1,155 @@
...
+func NewWalReader(limiter *ReadLimiter, reader io.ReaderAt, baseFileName string) (*WalReader, error) {
+	header := make([]byte, WalHeaderSize)
+	n, err := limiter.ReadAt(reader, header, 0)
+	if err != nil && !errors.Is(err, io.EOF) {
+		return nil, fmt.Errorf("failed to read WAL header: %w", err)
+	}
+	if n < WalHeaderSize {
+		return nil, fmt.Errorf("WAL file too short: expected at least %d bytes, got %d", WalHeaderSize, n)
+	}
+	magic := binary.LittleEndian.Uint32(header[0:unsafe.Sizeof(WalMagic)])
```
eguguchkin, line 39: nit: What's the plan if we find that WalHeader is corrupted?

storage/wal_reader.go
eguguchkin: The logic for reading and detecting bad blocks is great.

However, the tracking of bad blocks and logging are implemented inaccurately, which can be misleading during debugging.

For example, if the last two blocks are corrupted, then in the penultimate block the offset will be aligned, and an error will occur in the last block. In the logs, the "to" value will exceed the file size.

storage/wal_reader.go
eguguchkin: I mean that right now the service will crash and we won't be able to exec sh into the pod to move such wal to, say, `/data/.broken/` and restart the service.

We also need some metric indicator to track that we have corrupted wals, so we can set up an alert for them.

Yes, that’s right — it should be like this: we always move every corrupted WAL file to the `.broken` directory (regardless of whether we failed to recover the entire file or just lost a single block). We also create a special metric that shows the number of files in .broken. Then we can set up an alert if this value is >0.

We also need to consider whether we should copy the corresponding .doc file to .broken as well, because if needed, we could extract something useful from it too.

So, essentially, a new task is emerging here :)

cheb0: yep, sounds like a different PR. In any case, corrupted header is like million times less probable than corrupted block

storage/wal_reader.go
cheb0: made it `logCorruptionEnd(offset + n)`

cheb0: `which can be misleading during debugging`

in case of problems a developer still end up manually debugging this loop and log messages may be disregarded

storage/wal_reader.go
cheb0: fixed

storage/wal_reader.go
cheb0: right, if it's `read` syscall err, then we can just bypass error without corruption tracking

storage/wal_writer.go
cheb0: fixed

frac/file_writer_test.go
cheb0: deleted `frac/file_writer_test` since file writer now resides in storage package

frac/active.go
cheb0: fixed

frac/active.go
cheb0: fixed

frac/active.go
cheb0: for any fraction either `.wal` or `.meta` file exists, but they can't exist both at the same time. If it's `.meta` file, then it will be forever `.meta` even after numerous crashes.

storage/wal_writer.go
cheb0: fixed

PR #351, opened 2026-02-11
feat: add support for failing partial response

### LINKED ISSUE #350: Add support for partial response failure
The default behaviour of cluster version is the when some of the shards are inaccessible, seq-proxy still tries to return the search results from available shards with "partial response" error

This feature should allow to toggle this behaviour
1. By default, seq-db returns search results if some shards are unavailable (current behaviour), can be changed in config.
2. If search request comes with header `fail-partial-response=true/false`, the behaviour is overriden by it.

moflotas:
Fixes #350 

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/cmd/seq-db/seq-db.go
+++ b/cmd/seq-db/seq-db.go
@@ -273,6 +273,7 @@ func startStore(
 				Search: frac.SearchConfig{
 					AggLimits: frac.AggLimits{
 						MaxFieldTokens:     cfg.Limits.Aggregation.FieldTokens,
+						MaxFieldValues:     cfg.Limits.Aggregation.FieldValues,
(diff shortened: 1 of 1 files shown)


cmd/test/main.go
```diff
@@ -0,0 +1,21 @@
...
+
+func main() {
+	ch := make(chan int)
+	ctx := context.Background()
+	go func() {
+		<-ch
+	}()
+	select {
+	case <-ctx.Done():
+		fmt.Println("hui")
```
cheb0: so, which one is printed first?

cheb0 (approved):

PR #360, opened 2026-02-20
feat(query-engine): prototype streaming query engine

forshev:
## Summary

Introduces the foundation for a new **streaming search engine** that will enable searching by non-indexed fields.

### Key Changes
- New `Query Engine` with streaming `RecordProducer` interface
- SeqQL parser extensions: `filter`, `sort`, `limit`, `stats`, `fields` pipes
- gRPC streaming API for one-phase search (store side)

### Architecture
Instead of the traditional two-phase approach (find IDs → fetch documents), this engine streams documents directly through a chain of executors (filter → sort → limit → project), enabling efficient search on non-indexed fields.

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```


tests/integration_tests/integration_test.go
```diff
@@ -1840,3 +1840,111 @@ func (s *IntegrationTestSuite) TestSkipMaskManager() {
...
+	t := s.T()
+	r := require.New(t)
+
+	cfg := *s.Config
+	cfg.EnableOnePhaseSearch = true
+	env := setup.NewTestingEnv(&cfg)
+
+	getNextTs := getAutoTsGenerator(time.Now(), -time.Nanosecond)
+
+	toBulk := []string{
```
cheb0, line 1854: if it's a single bulk, then we don' test in a cluster env?

api/storeapi/store_api.proto
```diff
@@ -264,3 +268,60 @@ message StatusRequest {}
...
+message OnePhaseSearchRequest {
+  string query = 1;
+  google.protobuf.Timestamp from = 2;
+  google.protobuf.Timestamp to = 3;
+  int64 size = 4;
+  int64 offset = 5;
+  bool explain = 6;
+  bool with_total = 7;
+  Order order = 8;
+  string offset_id = 9;
```
cheb0, line 281: TODO: `offset_id` not supported

api/storeapi/store_api.proto
```diff
@@ -264,3 +268,60 @@ message StatusRequest {}
 message StatusResponse {
   google.protobuf.Timestamp oldest_time = 1;
 }
+
+message OnePhaseSearchRequest {
+  string query = 1;
+  google.protobuf.Timestamp from = 2;
+  google.protobuf.Timestamp to = 3;
+  int64 size = 4;
+  int64 offset = 5;
```
cheb0, line 277: TODO: we need to establish the API for query engine. Currently, I can set both `limit` and `size` and they work both at the same time, which is odd.

tests/integration_tests/integration_test.go
```diff
@@ -1840,3 +1840,111 @@ func (s *IntegrationTestSuite) TestSkipMaskManager() {
...
+		fmt.Sprintf(`{"ts":%q, "level":3, "service":"service1", "unindexed":"1", "sorted":"1"}`, getNextTs()),
```
cheb0, line 1855: I would like to match the behaviour of existing search API. Like `service:travel | limit 100` must match ordinary search with `size` set to 100 with both orders. Ideally both on low and large (thousands) amount of docs.

Currently, I'm running a cluster env and unable to get same results using both old and new API for `limit`.

query/exec/data_source_test.go
```diff
@@ -0,0 +1,135 @@
...
+		`{"timestamp":"2000-01-01T13:00:00.004Z","service":"service3", "level":3}`,
+		`{"timestamp":"2000-01-01T13:00:00.005Z","service":"service3", "level":3}`,
+		`{"timestamp":"2000-01-01T13:00:00.006Z","service":"service4", "level":3}`,
+		`{"timestamp":"2000-01-01T13:00:00.007Z","service":"service4", "level":5}`,
+		`{"timestamp":"2000-01-01T13:00:00.008Z","service":"service5", "level":3}`,
+		`{"timestamp":"2000-01-01T13:00:00.009Z","service":"service5", "level":3}`,
+	}
+	fraction := setupFraction(t, mapping, docs)
+
+	queryString := "level:in(3, 5)"
```
cheb0, line 49: I think it looks like something we would put to `fraction_test`. We could add tests there (or make a new test) where we would test different expressions written in a new language against different fractions.

However, we will still tests most stuff in integration tests because cluster env tests are a must.

proxy/search/one_phase_search.go
```diff
@@ -0,0 +1,293 @@
...
+		default:
+			// unexpected error on all hot replica sets (usually bad query)
+			return nil, err
+		}
+	}
+
+	var mergedStream query.RecordProducer
+
+	hasAggs := len(sr.AggQ) > 0 // TODO: check based on stats pipe
+	if hasAggs {
```
cheb0, line 70: TODO: 'stats' pipe ignores every pipe written after it (sort/limit/filter)

UPD (probably most critical issue): I would actually generalize that we have no query processing on proxy level here. Generally speaking, most queries I would write will not work here. If they do not work, then such queries should not be allowed at all. For example, the query  `service:payments | sort amount desc | limit 10` will never work - we need sort amongst all logs which we decided we will not implement. However, currently there is no problem in executing such a query and it produces inconsistent results. Any user would expect we will get top 10 logs by amount among all logs in a cluster while issuing a query `service:payments | sort amount desc | limit 10`. However, it doesn't work that way.

The primary goal of this PR would be to come up with robust set of queries which work and produce consistent queries. Unsupported queries should be disallowed.

tests/integration_tests/integration_test.go
cheb0, line 1855: let's shuffle `sorted` field values so that sorter sorts something

query/exec/limiter.go
```diff
@@ -0,0 +1,42 @@
+package exec
+
+import "github.com/ozontech/seq-db/query"
+
+type Limiter struct {
```
cheb0, line 5: Please explain how to make limiter match the current API.

I do two requests:

```
curl --request POST --url http://localhost:9102/complex-search --header 'Content-Type: application/json' \
  --data-binary '
  {
    "query":{
      "query":"service:payment*",
      "from": "2026-07-10T02:24:37.456Z",
      "to": "2026-07-10T02:25:00.218Z",
      "explain": false
    },
    "size": 100,
    "order": "ORDER_ASC"
  }' | jq
```

and 

```
curl --request POST --url http://localhost:9102/one-phase-search --header 'Content-Type: application/json' \
  --data-binary '
  {
    "query":{
      "query":"service:payment* | limit 100",
      "from": "2026-07-10T02:24:37.456Z",
      "to": "2026-07-10T02:25:00.218Z",
      "explain": false
    },
    "size": 100
  }' | jq
```

I'm always get different results. I tried both `asc` and `desc` order.

api/storeapi/store_api.proto
```diff
@@ -264,3 +268,60 @@ message StatusRequest {}
...
+  FLOAT64 = 7;
+  // TODO: array data types: StringArray, Uin64Array, Float64Array etc.
+}
+
+message Typing {
+  string title = 1;
+  DataType type = 2;
+}
+
+message RecordsBatch {
```
cheb0, line 321: TODO: columnar batches will not aligh will with that model. But currently it's ok. Probably batch format will be just raw bytes soon.

query/exec/sorter.go
```diff
@@ -0,0 +1,112 @@
...
+
+	insaneJSON "github.com/ozontech/insane-json"
+
+	"github.com/ozontech/seq-db/query"
+)
+
+// TODO: use existing type ???
+type Order byte
+
+const (
```
cheb0, line 14: Sorting works not as I would expect. 

`service:payment* | sort time asc | limit 1000` doesn't use order asc in `SearchParams` and sorts all data within the specified limit.

Basically it does exactly what `service:payment* | limit 1000 | sort time asc` I would expect to do.

query/exec/filter.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"cmp"
+
+	insaneJSON "github.com/ozontech/insane-json"
+
+	"github.com/ozontech/seq-db/query"
+)
+
+type FilterExpr[T any] interface {
```
cheb0, line 11: filtering like that `service:payment* | filter level:5 | limit 100` produces odd results, because filter is not pushed to inverted index. I remember we have discussed it. 

It look like a problem to me simple because a user doesn't know if field is indexed or not, i.e. 

I think that queries `service:payment* AND level:5 | limit 100` AND `service:payment* | filter level:5 | limit 100`  must produce same result, but they clearly don't!

First query will search all level:5 in a large timeframe and will return 100 docs. Second query will search for 100 docs of `service:payment*` and look for `level:5` among them. Second query can potentially yield zero docs! I would even argue field pipe doesn't work as it should.

storeapi/grpc_one_phase_search.go
```diff
@@ -0,0 +1,351 @@
...
+
+	if len(statsAggs) > 0 {
+		// TODO: combine aggs with another executors
+		return producer, nil
+	}
+
+	for _, pipe := range seqql.Pipes {
+		switch p := pipe.(type) {
+		case *parser.PipeFilter:
+			producer = exec.NewFilter(producer, 2, exec.NewDocFilter(p.Condition.Field, exec.NewEq(p.Condition.Value)))
```
cheb0, line 228: hardcoded column index

storeapi/grpc_one_phase_search.go
```diff
@@ -0,0 +1,351 @@
...
+		case *parser.PipeSort:
+			order := exec.OrderAsc
+			if p.Order == "desc" {
+				order = exec.OrderDesc
+			}
+			producer = exec.NewDocSorter(producer, 2, p.Field, order)
```
cheb0, line 234: same, hardcoded column index

query/exec/data_source.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/ozontech/seq-db/frac"
+	"github.com/ozontech/seq-db/frac/processor"
+	"github.com/ozontech/seq-db/fracmanager"
+	"github.com/ozontech/seq-db/pkg/storeapi"
+	"github.com/ozontech/seq-db/query"
+	"github.com/ozontech/seq-db/querytracer"
+	"github.com/ozontech/seq-db/seq"
+)
+
+type FractionDataSource struct {
```
cheb0, line 20: note (for other reviewers): is not used anywhere other than `data_source_test.go`

query/producer.go
```diff
@@ -0,0 +1,15 @@
...
+	Next() (*Record, *Metadata)
+	Release()
+}
+
+type Metadata struct {
+	Err error
+	// TODO: some additional info like explain data, tracing data, etc
+}
+
+type BatchedRecordProducer interface {
```
cheb0, line 13: note: is not used anywhere

consts/consts.go
```diff
@@ -111,4 +111,5 @@ var (
 	ErrTooManyFractionTokens     = errors.New("aggregation has too many fraction tokens")
 	ErrTooManyFractionsHit       = errors.New("too many fractions hit")
 	ErrMemoryLimitExceeded       = errors.New("memory limit exceeded")
+	ErrResourceDisabled          = errors.New("resource is disabled")
```
cheb0, line 114: what's the concrete meaning of `resource` here?

config/config.go
```diff
@@ -300,6 +300,8 @@ type Config struct {
 		// Specify how many tokens can be checked using regular expressions.
 		// If zero then there is no limit.
 		MaxRegexTokensCheck int `config:"max_regex_tokens_check" default:"0"`
+		// EnableOnePhaseSearch enanles experimental one phase search method that supports new query engine.
+		EnableOnePhaseSearch bool `config:"enable_one_phase_search"`
```
cheb0, line 304: Even with disabled toggle I can issue a new query `service:payment* AND level:5 | sort level asc | limit 100` using complex search and I do not have parse error, but I think I should.

api/seqproxyapi/v1/seq_proxy_api.proto
```diff
@@ -111,6 +111,13 @@ service SeqProxyApi {
       body: "*"
     };
   }
+
+  rpc OnePhaseSearch(SearchRequest) returns (ComplexSearchResponse) {
```
cheb0, line 115: Duplicating an issue here - currently we must set `size` even in presence of `limit` pipe.

api/storeapi/store_api.proto
cheb0, line 277: TODO: `offset` is NOP

query/exec/merger.go
```diff
@@ -0,0 +1,182 @@
...
+
+	"github.com/ozontech/seq-db/query"
+)
+
+type Merger struct {
+	left, right query.RecordProducer
+
+	curLeft, curRight *query.Record
+
+	colIdx   int
```
cheb0, line 16: since `colIdx` is exact same for all docs, how do we guarantee all docs have field, let's say `amount`, on same column index?

PR #364, opened 2026-02-27
feat: implement regions

moflotas:
# Description

Fixes #363

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```


config/config.go
```diff
@@ -281,6 +281,19 @@ type Config struct {
 		// Specify how many tokens can be checked using regular expressions.
 		// If zero then there is no limit.
 		MaxRegexTokensCheck int `config:"max_regex_tokens_check" default:"0"`
+
+		// Regions configures the proxy to query only remote regions (other seq-proxies via store API).
+		// When set, hot_stores, hot_read_stores, write_stores, read_stores must be empty. No bulk; search only.
+		Regions struct {
+			// UseRegions is whether to use regions or not
+			UseRegions bool `config:"use_regions"`
```
cheb0, line 289: currently, config looks like that:
```
experimental:
    regions:
        use_regions: true
```

We repeat `regions` twice here.

I think it would be cleaner to have just:
```
experimental:
    regions:
        enabled: true
```

proxyapi/store_emulator.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/ozontech/seq-db/proxy/search"
+	"github.com/ozontech/seq-db/querytracer"
+	"github.com/ozontech/seq-db/storage"
+	store "github.com/ozontech/seq-db/storeapi"
+	"github.com/ozontech/seq-db/util"
+)
+
+// storeEmulator implements store API (Search, Fetch, Status) so seq-proxy can be
+// queried as a store for aggregation-friendly cross-region queries (e.g. quantiles).
+type storeEmulator struct {
```
cheb0, line 23: For me `storeAdapter` would be a better name.

proxyapi/store_emulator.go
```diff
@@ -0,0 +1,105 @@
...
+	storeapi.UnimplementedStoreApiServer
+	searchIngestor SearchIngestor
+}
+
+func newStoreEmulator(si SearchIngestor) *storeEmulator {
+	return &storeEmulator{searchIngestor: si}
+}
+
+func (s *storeEmulator) Search(ctx context.Context, req *storeapi.SearchRequest) (*storeapi.SearchResponse, error) {
```
cheb0, line 32: One thing: `explain` doesn't work for regions. I set up a simple region cluster and here are the results:

Explained query for regions:
```
{
  "error": {
    "code": "ERROR_CODE_NO",
    "message": ""
  },
  "explain": {
    "duration": "316.815518ms",
    "message": "proxy/ComplexSearch",
    "children": [
      {
        "duration": "206.588436ms",
        "message": "proxy/searchShard",
        "children": [
          {
            "duration": "17.383µs",
            "message": "Making search request to localhost:9114"
          }
        ]
      },
      {
        "duration": "316.62144ms",
        "message": "proxy/searchShard",
        "children": [
          {
            "duration": "48.615µs",
            "message": "Making search request to localhost:9124"
          }
        ]
      }
    ]
  }
}
``` 

Explained query for a single region:

```
{
  "error": {
    "code": "ERROR_CODE_NO",
    "message": ""
  },
  "explain": {
    "duration": "1.5405ms",
    "message": "proxy/ComplexSearch",
    "children": [
      {
        "duration": "1.41697ms",
        "message": "proxy/searchShard",
        "children": [
          {
            "duration": "33.876µs",
            "message": "Making search request to localhost:9014"
          },
          {
            "duration": "157.948µs",
            "message": "store/Search",
            "children": [
              {
                "duration": "17.734µs",
                "message": "parse query"
              },
              {
                "duration": "8.126µs",
                "message": "search iteratively"
              },
              {
                "duration": "0s",
                "message": "search iteratively (cpu time)"
              },
              {
                "duration": "0s",
                "message": "waiting goroutines (all cores)"
              }
            ]
          }
        ]
      }
    ]
  }
}

```

config/config.go
```diff
@@ -281,6 +281,19 @@ type Config struct {
...
+			// Regions maps region name to store API gRPC address (e.g. "eu" -> "eu-proxy:9004").
+			Regions map[string]string `config:"regions"`
+			// MaxConcurrent limits how many regions are queried in parallel per search (0 = no limit).
+			MaxConcurrent int `config:"max_concurrent" default:"0"`
```
cheb0, line 293: I think it's hard to tell what `max_concurrent` means without reading a code comment.
```
experimental:
  regions:
    use_regions: true
    ...
    max_concurrent: 5
```

proxyapi/store_emulator.go
```diff
@@ -0,0 +1,105 @@
...
+	sr := search.SearchRequestFromStoreAPI(req)
```
cheb0, line 33: Store adapter doesn't bypass store protocol version. It means regions proxy currently get store protocol version as `2` while the primary proxy gets `1`. It's possible that proxy can misinterpret QPRs from regions.

pkg/seqproxyapi/v1/seq_proxy_api.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.36.5
-// 	protoc        v5.29.3
+// 	protoc        v6.32.1
```
cheb0, line 4: Are we sure it's safe change this? I assume you used newer `protoc` to generate proto files. I'm personally not sure if we want to this in this PR.

config/config.go
```diff
@@ -281,6 +281,19 @@ type Config struct {
 		// Specify how many tokens can be checked using regular expressions.
 		// If zero then there is no limit.
 		MaxRegexTokensCheck int `config:"max_regex_tokens_check" default:"0"`
+
+		// Regions configures the proxy to query only remote regions (other seq-proxies via store API).
+		// When set, hot_stores, hot_read_stores, write_stores, read_stores must be empty. No bulk; search only.
```
cheb0, line 286: nit: Even with proxy setup I still can configure hot_stores:

```
cluster:
  replicas: 1
  hot_stores:
    - localhost:9004
    - localhost:9008
```

However, I can't use them:

```
{
  "code": 3,
  "message": "no regions specified in request (required when using experimental.regions)"
}
```

I guess, we could validate config that `hot_stores` can't be set if regions are enabled.

proxy/search/ingestor.go
```diff
@@ -33,6 +34,11 @@ type Config struct {
 	WriteStores     *stores.Stores
 	ShuffleReplicas bool
 	MirrorAddr      string
+
+	// RegionStores is a separate topology for experimental.regions (not mixed with hot_stores). When set, search uses only this.
+	RegionStores *stores.Stores
+	// RegionNames maps shard index to region name (same order as RegionStores.Shards).
+	RegionNames []string
```
cheb0, line 41: nit: so we basically have region names decoupled from region stores, but indexes are same. This is a bit ugly. I suppose it's not possible to solve this problem altogether, but I would not be against moving `regions` inside `stores.Stores`. Default setup could have empty (not used) region names

proxy/search/ingestor.go
```diff
@@ -82,16 +88,41 @@ func (si *Ingestor) Search(
 	}
 
 	startTime := time.Now()
-	searchStores := si.config.HotStores
-	if si.config.HotReadStores != nil && len(si.config.HotReadStores.Shards) > 0 {
-		searchStores = si.config.HotReadStores
+
+	var searchStores *stores.Stores
+	if config.UseRegions {
```
cheb0, line 93: nit: I'm not against having longer methods. However, here I would really move this logic of determing which stores we call to some method like `locateStores` or something like that.

proxy/search/ingestor.go
```diff
@@ -82,16 +88,41 @@ func (si *Ingestor) Search(
...
+		// Regions-only path: use RegionStores only; hot_stores logic is untouched.
+		selected := sr.Regions
+		allowed := make(map[string]struct{}, len(selected))
+		for _, n := range selected {
+			allowed[strings.TrimSpace(n)] = struct{}{}
+		}
+		shards := make([][]string, 0)
+		vers := make([]string, 0)
+		for i, name := range si.config.RegionNames {
+			if _, ok := allowed[name]; ok && i < len(si.config.RegionStores.Shards) {
```
cheb0, line 103: nit (not really): would be cool to support wildcards in future. Users could specify regions like `krd*`

proxyapi/store_emulator.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/ozontech/seq-db/querytracer"
+	"github.com/ozontech/seq-db/storage"
+	store "github.com/ozontech/seq-db/storeapi"
+	"github.com/ozontech/seq-db/util"
+)
+
+// storeEmulator implements store API (Search, Fetch, Status) so seq-proxy can be
+// queried as a store for aggregation-friendly cross-region queries (e.g. quantiles).
+type storeEmulator struct {
+	storeapi.UnimplementedStoreApiServer
```
cheb0, line 24: nit: I presume, we do not support async searches with regions enabled. It would be reasonable if we got some errors clarifying that. Currently, I get: `can't find store shards in config` when async search request is issued.

cheb0 (changes requested):

pkg/seqproxyapi/v1/seq_proxy_api.pb.go
eguguchkin: Good point! But as far as I understand, this refers to the compiler version upgrade, not the protobuf binary format

config/config.go
```diff
@@ -281,6 +281,19 @@ type Config struct {
...
+		Regions struct {
+			// UseRegions is whether to use regions or not
+			UseRegions bool `config:"use_regions"`
+			// Regions maps region name to store API gRPC address (e.g. "eu" -> "eu-proxy:9004").
+			Regions map[string]string `config:"regions"`
+			// MaxConcurrent limits how many regions are queried in parallel per search (0 = no limit).
+			MaxConcurrent int `config:"max_concurrent" default:"0"`
+			// MaxRegionsPerRequest is the maximum number of regions a client can specify in one search request (e.g. 5).
+			MaxRegionsPerRequest int `config:"max_regions_per_request" default:"5"`
```
eguguchkin, line 295: nit: Honestly, I would just remove these Max* options altogether. Yes, we'll have 300+ regions, but it seems like a manageable task to merge the results, especially once we have streaming single-phase search.

proxy/search/store_response.go
```diff
@@ -0,0 +1,75 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("wrong doc id %s format: %w", s, err)
+		}
+		ids = append(ids, parsed)
+	}
+	return ids, nil
+}
+
+// SearchRequestFromStoreAPI converts store API SearchRequest to proxy search SearchRequest.
+func SearchRequestFromStoreAPI(req *storeapi.SearchRequest) *SearchRequest {
```
eguguchkin: раз мы вынесли в отдельную функцию давай использовать во всех местах где мы получаем SearchRequest

cmd/seq-db/seq-db.go
```diff
@@ -219,7 +194,66 @@ func startProxy(
 		},
 	}
 
-	ingestor, err := proxyapi.NewIngestor(pconfig, inMemory)
+	regionsCfg := cfg.Experimental.Regions
+	// without regions
+	if !config.UseRegions {
```
eguguchkin, line 199: nit: For better readability, let's just extract this setup logic into a separate function, which in turn will call one of two functions - for the experimental config or for the traditional one.

```go
func setupStores(cfg config.Config, pConfig proxyapi.IngestorConfig) proxyapi.IngestorConfig {
	if !cfg.Experimental.Regions.UseRegions {
		setupRegionStores(cfg, pConfig)
	}
	return setupTraditionalStores(cfg, pConfig)
}
```

config/shared.go
```diff
@@ -17,4 +17,6 @@ var (
 	MaxRegexTokensCheck int
 
 	FailPartialResponse = false
+
+	UseRegions = false
```
eguguchkin, line 21: nit: I would avoid creating yet another global config variable. They create complications down the road (leaky abstraction).

Everywhere we use this variable, there is indirect data in the context about whether regions are enabled or not, so we can easily do without this variable.

proxyapi/store_emulator.go
```diff
@@ -0,0 +1,105 @@
...
+		if err := stream.Send(&storeapi.BinaryData{Data: block}); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// TODO(moflotas) in case of regions work improperly
+func (s *storeEmulator) Status(_ context.Context, _ *storeapi.StatusRequest) (*storeapi.StatusResponse, error) {
+	st := s.searchIngestor.Status(context.Background())
```
eguguchkin, line 79: Why is a new context being created instead of passing it from the parameters?

proxy/search/ingestor.go
eguguchkin: In my opinion, there are quite a few ugly spots in the proxy design right now.

This PR is the first prototype phase of the solution. We plan to introduce unified (non‑experimental) configs for describing cluster topology with region support.

I think it would be good, during the second phase, to also clean up the structures and entities in this part of the project.

config/config.go
eguguchkin: nit: Honestly, I would just remove these Max* options altogether. Yes, we'll have 300+ regions, but it seems like a manageable task to merge the results, especially once we have streaming single-phase search.

proxy/search/ingestor.go
eguguchkin: This is a very, very basic implementation. It's a good idea for a future enhancement.

Issue #336, opened 2026-01-30
Implement compaction of sealed fractions


dkharms:
As an important milestone for seq-db, we plan to introduce compaction into the database.

We expect several benefits from compaction:
- It will allow us to reduce the fraction size to very small values (or even seal a fraction for each bulk), significantly lowering the memory footprint. We also anticipate an increase in ingestion throughput, since smaller fractions should reduce contention;
- We expect lower on-disk usage for fractions. For example, the tokens section of the `.index` file (which accounts for around 20% of a fraction’s `.index` size) often contains overlapping `(field, value)` tuples across multiple fractions. When fractions are merged, these sections become more space-efficient to store by merging duplicate tuples;
- Compaction will enable us to implement partitioning. We expect partitions (and parts) to be created upon sealing, which may produce many tiny sealed fractions — an inefficient outcome. Background compaction will address this by merging small parts into larger, more efficient ones;
- Also, if we stick with Time-Tiered Compaction Strategy (TWCS, DTCS) we can easily implement time-based retention.

PR #389, opened 2026-03-31
refactor: split `.index` file into multiple

dkharms:
# Description

This is first pull request of series #336. Changes in this pull request will allow us to efficiently merge several fractions into one when performing compaction.

So `.index` gets split into several:
- `.info` -- contains one info block;
- `.offsets` -- contains one block with offsets of `DocBlock` inside `.docs` file;
- `.ids` -- contains triplets of `seq.MID`, `seq.RID` and `seq.DocPos` blocks;
- `.tokens` -- contains tokens and token table;
- `.lids` -- contains chunks of `seq.LID`;

It's easy to notice that we expect increase in file descriptors usage by 3x

Several things I need to polish:
- [x] Reintroduce statistics reporting on sealing;
- [x] Backwards compatibility for sealed fractions which were offloaded to remote storage;
- [x] Delete all sealed fractions files if at least one file has tmp suffix;
- [x] Handle tmp `.index` files; 

---


dkharms:
@seqbenchbot up main bulk

dkharms:
@seqbenchbot down f70e7779

dkharms:
@seqbenchbot up main mixed

dkharms:
@seqbenchbot down e47604c9

dkharms:
@seqbenchbot up main bulk

dkharms:
@seqbenchbot down 3a13f7af

dkharms:
@seqbenchbot up main bulk

dkharms:
@seqbenchbot down 29e733d8

dkharms:
@seqbenchbot up main bulk

dkharms:
@seqbenchbot down b1727103

cheb0 (approved):
looks good to me

dkharms:
@seqbenchbot up main bulk

dkharms:
@seqbenchbot down cfc11182

dkharms:
@seqbenchbot up main bulk

dkharms:
@seqbenchbot down 3dd6ec65

dkharms:
@seqbenchbot up main bulk

dkharms:
@seqbenchbot down ddbdf38d

dkharms:
@seqbenchbot up main bulk

dkharms:
@seqbenchbot down 12d95a0e

dkharms:
@seqbenchbot up main bulk

dkharms:
@seqbenchbot down d1043cea

dkharms:

dkharms:
@seqbenchbot down caad43de

dkharms:
@seqbenchbot up main mixed

dkharms:
@seqbenchbot down 1d4eb6bd

frac/sealed/sealing/blocks_builder.go
```diff
@@ -45,94 +49,110 @@ type idsSealBlock struct {
 
 // blocksBuilder constructs sealed blocks from various data sources.
 // Provides error tracking and consistency validation during block construction.
-type blocksBuilder struct {
-	lastErr error // Last error encountered during processing
-}
+type blocksBuilder struct{}
```
eguguchkin, line 52: Do we still need this structure now? We can now simply use the BuildTokenBlocks function. Also, this comment is outdated.

frac/sealed/sealing/blocks_builder.go
```diff
@@ -45,94 +49,110 @@ type idsSealBlock struct {
...
-// So we split field ranges into field entries - sub-ranges of fields aligned to block boundaries.
-// Each field table (token.FieldTable) links a field to a blocks and token ranges inside the blocks.
-//
-// Parameters:
-//   - tokenBatches: Iterator of token batches, where each batch becomes a separate block
-//   - fields: Iterator of [fieldName, maxTID] pairs for all fields in ascending TID order
-//
-// Returns: Iterator of [token block, field table for block] pairs, where field table contains
-// information about which fields and their ranges are represented in this block.
 func (bb *blocksBuilder) BuildTokenBlocks(
```
eguguchkin, line 54: nit: Maybe we shouldn't extract the lead accumulation logic but hide it inside the function? It seems like it will always be the same - splitting into blocks. The only thing that might vary is the block processing logic (onBlock). Maybe we could pass an onBlock func instead of an accumulate func?

```diff
diff --git a/frac/sealed/sealing/blocks_builder.go b/frac/sealed/sealing/blocks_builder.go
index 6c295903..086ebb28 100644
--- a/frac/sealed/sealing/blocks_builder.go
+++ b/frac/sealed/sealing/blocks_builder.go
@@ -53,7 +53,7 @@ type blocksBuilder struct{}

 func (bb *blocksBuilder) BuildTokenBlocks(
        it iter.Seq2[string, iter.Seq2[TokenPosting, error]],
-       accumulate func([]uint32) error, blockCapacity int,
+       onBlock func(lidsSealBlock) error, tokenBlockCapacity, lidsBlockCapacity int,
 ) iter.Seq2[TokenBlock, error] {
        return func(yield func(TokenBlock, error) bool) {
                var (
```

However, that would mean we also need to pass the lead block size to the builder…

eguguchkin, line 54: nit: I'd also like some comment on the function. You could reference TokenTriplet to make the first parameter clearer, and duplicate what is already written in the comment for TokenTriplet.

And regarding the old comment with the diagram. Do you think the old visualization didn't help to understand the resulting index structure?

frac/sealed/sealing/blocks_builder.go
```diff
@@ -45,94 +49,110 @@ type idsSealBlock struct {
...
+			blockSize = 0
+
+			pendingTable = pendingTable[:0]
+			fieldEntryStartTID = currentTID + 1
+
+			return true
+		}
+
+		block.ext.minTID = 1
+		for field, tokIt := range it {
```
eguguchkin: nit: tokIt isn't a very expressive abbreviation in my opinion. I'd prefer something like 
```go
for field, tokenIterator := range tokenTriplet {
```

frac/sealed/sealing/blocks_builder.go
eguguchkin, line 54: nit: This won't change the logic, but if we now move the local variables from the closure into struct fields, the code might become more readable. Something like:

```go
type blockBuilder struct {
	// External dependencies
	maxBlockSizeBytes int                          // Maximum block size in bytes
	lidsCollector     func([]uint32) error         // Function to accumulate token IDs
	yieldBlock        func(TokenBlock, error) bool // Function to return the result

	// Current block state
	blockIndex       uint32          // Sequence number of current block
	currentBlock     tokensSealBlock // Block being assembled
	currentBlockSize int             // Current block size in bytes
	currentTID       uint32          // Last assigned token ID

	// Field processing state
	currentFieldName   string             // Name of field being processed
	pendingFieldTables []token.FieldTable // Accumulated field table entries
	fieldEntryStartTID uint32             // First token ID of current field entry
}
```

frac/sealed/sealing/blocks_builder.go
```diff
@@ -148,159 +168,111 @@ func createTokenTableEntry(entryStartTID, fieldMaxTID, blockIndex uint32, block
...
-			isEndOfToken = true
+		if len(block.mids.Values) > 0 {
+			yield(block, nil)
 		}
+	}
+}
 
-		// Yield the final block
-		finalizeBlock()
+type lidBlocksAcc struct {
```
eguguchkin: nit: I don't really like an abbreviation like Acc. Maybe `lidsBlockCollector` would fit better here?

frac/sealed/sealing/blocks_builder.go
```diff
@@ -148,159 +168,111 @@ func createTokenTableEntry(entryStartTID, fieldMaxTID, blockIndex uint32, block
...
-
-			// Convert each ID and position to storage format
-			for i, id := range ids {
-				block.mids.Values = append(block.mids.Values, uint64(id.MID))
-				block.rids.Values = append(block.rids.Values, uint64(id.RID))
-				block.params.Values = append(block.params.Values, uint64(positions[i]))
+// For each block that fills up, `onBlock` is called immediately
+// before the backing arrays are reset, so `onBlock` may read the
+// block data but must not retain references to it.
+func (a *lidBlocksAcc) Add(lidsbuf []uint32, onBlock func(lidsSealBlock) error) error {
```
eguguchkin: nit: Maybe pass onBlock to the constructor once?

util/pair.go
```diff
@@ -0,0 +1,6 @@
+package util
+
+type Pair[F, S any] struct {
```
eguguchkin, line 3: This is an unnecessary abstraction. Let's declare the corresponding types directly. Compare:

```go
	DocLocation struct {
		id  seq.ID
		pos seq.DocPos
	}
	TokenPosting struct {
		token []byte
		lids  []uint32
	}
```
etc.

It will be clearer and have fewer intermediate entities:

```diff
diff --git a/frac/sealed/sealing/blocks_builder.go b/frac/sealed/sealing/blocks_builder.go
index 6c295903..39034f44 100644
--- a/frac/sealed/sealing/blocks_builder.go
+++ b/frac/sealed/sealing/blocks_builder.go
@@ -174,16 +174,15 @@ func seqBlockID(ids iter.Seq2[DocLocation, error], blockCapacity int) iter.Seq2[
        return func(yield func(idsSealBlock, error) bool) {
                var block idsSealBlock

-               for pair, err := range ids {
+               for doc, err := range ids {
                        if err != nil {
                                yield(idsSealBlock{}, err)
                                return
                        }

-                       id, pos := pair.First, pair.Second
-                       block.mids.Values = append(block.mids.Values, uint64(id.MID))
-                       block.rids.Values = append(block.rids.Values, uint64(id.RID))
-                       block.params.Values = append(block.params.Values, uint64(pos))
+                       block.mids.Values = append(block.mids.Values, uint64(doc.id.MID))
+                       block.rids.Values = append(block.rids.Values, uint64(doc.id.RID))
+                       block.params.Values = append(block.params.Values, uint64(doc.pos))

                        if len(block.mids.Values) == blockCapacity {
                                if !yield(block, nil) {
```

fracmanager/loader.go
```diff
@@ -136,9 +137,23 @@ func (l *Loader) discover(ctx context.Context) ([]*frac.Active, []*frac.Sealed,
 		case fracStageActive:
 			actives = append(actives, l.provider.NewActive(manifest.basePath))
 		case fracStageSealed:
-			locals = append(locals, l.loadSealed(manifest.basePath, loadedInfoCache))
+			locals = append(locals, l.loadSealed(manifest, loadedInfoCache))
 		case fracStageRemote:
-			remotes = append(remotes, l.loadRemote(ctx, manifest.basePath, loadedInfoCache))
+			// TODO(dkharms): Drop this compatibility check.
```
eguguchkin: This looks like a workaround. Do you have an understanding of how to solve this properly in the future?

By the way, it makes sense to bump the fraction version stored in info, and then determine whether it's legacy or not from info itself.

fracmanager/frac_manifest.go
```diff
@@ -88,13 +110,13 @@ func (m *fracManifest) Stage() fracStage {
 	if m.hasRemote {
 		return fracStageRemote
 	}
-	if m.hasIndex && (m.hasSdocs || m.hasDocs) {
+	if (m.hasAllIndexFiles() || m.hasIndex) && (m.hasSdocs || m.hasDocs) {
```
eguguchkin, line 113: nit: Right now `m.hasAllIndexFiles()` is always checked together with `m.hasIndex`. If you don't plan to check it separately from `m.hasAllIndexFiles()` in the future, you can move the `m.hasIndex` check inside the `hasAllIndexFiles` function.

frac/index_cache.go
```diff
@@ -8,7 +8,14 @@ import (
 )
 
 type IndexCache struct {
-	Registry   *cache.Cache[[]byte]
+	// Per-file registry caches (each IndexReader needs its own).
+	InfoRegistry    *cache.Cache[[]byte]
```
eguguchkin: I already mentioned that the registry itself is redundant for info, but even if we kept it, the cache for it wouldn't be needed, because info is read only once.

seq/seq.go
```diff
@@ -10,14 +10,23 @@ import (
 	"github.com/ozontech/seq-db/util"
 )
 
+var (
+	SystemMID    MID    = math.MaxUint64
+	SystemRID    RID    = math.MaxUint64
+	SystemID     ID     = ID{SystemMID, SystemRID}
+	SystemDocPos DocPos = DocPos(0)
```
eguguchkin, line 17: nit: I would declare all these systemID locally (as they were), because these are implementation details — or more precisely, details of data representation on disk (sealed) or in memory (active) — and for other implementations, these entities may not be relevant.

In other words, in the general concept of fractions, system IDs don't exist. They only exist as stubs in a specific implementation.

frac/sealed/sealing/writer.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/ozontech/seq-db/bytespool"
+	"github.com/ozontech/seq-db/logger"
+	"github.com/ozontech/seq-db/util"
+)
+
+const prefixSize = 16
+
+const (
+	btypeInfo       = "info"
```
eguguchkin: nit: My IDE highlights this as a typo because it doesn't recognize the word btype.
 - btypeInfo -> bTypeInfo?
 - btypeInfo -> blockTypeInfo?

frac/sealed/sealing/writer.go
```diff
@@ -0,0 +1,134 @@
...
+	btypeOffset     = "offset"
+	btypeToken      = "token"
+	btypeTokenTable = "token-table"
+	btypeMid        = "mid"
+	btypeRid        = "rid"
+	btypeDocPos     = "doc-pos"
+	btypeLid        = "lid"
+	btypeBlackhole  = "blackhole"
```
eguguchkin: nit: I'll also nitpick the naming here, because this isn't really a blackhole — it's an actual empty block in form and EndOfSection in meaning.

How about extracting this into a separate method?
`w.writeBlock(btypeBlackhole, indexBlock{}) => w.writeNullBlock()`

Then we won't need to extract the type into a constant either. But overall, having an empty block at the end is questionable for a sharded index.

frac/sealed/sealing/index.go
```diff
@@ -49,295 +46,203 @@ func NewIndexSealer(params common.SealParams) *IndexSealer {
...
 	}
+	defer lw.release()
+
+	var (
+		bb              blocksBuilder
+		allFieldsTables []token.FieldTable
+		lidacc          = newLIDBlocksAccumulator(consts.LIDBlockCap)
+	)
+
+	// NOTE(dkharms): This is so ugly but I cannot come up with other solution here.
```
eguguchkin: nit: Let's remove this comment and instead write something explanatory that will help the reader appreciate the beauty of the construct :)

Overall, here's what I can recommend:
- move `onBlock` to the constructor
- make a `lidacc.Finalize()` method (instead of `Flush`) and use the previously passed `onBlock` inside it
- accordingly, remove `s.finalizeLIDFile()` and call this instead:

```go
lidacc.Finalize()
w.writeBlock(btypeBlackhole, indexBlock{})
```

I'd suggest going further and reconsidering the need to insert an empty block at the end, because it's redundant. We can determine the last block when reading via the registry.

frac/sealed/sealing/index.go
```diff
@@ -49,295 +46,203 @@ func NewIndexSealer(params common.SealParams) *IndexSealer {
...
-		}
+	return w.finalize()
+}
 
-		if !sectionSeparator() {
-			return
-		}
+func (s *IndexSealer) finalizeTokenFile(w *writer, allFieldsTables []token.FieldTable) error {
+	// Emit section separator.
+	if err := w.writeBlock(btypeToken, indexBlock{}); err != nil {
```
eguguchkin: nit: Did you really intend to use this specific block type? Won't it break the statistics?

frac/sealed/sealing/index.go
```diff
@@ -49,295 +46,203 @@ func NewIndexSealer(params common.SealParams) *IndexSealer {
...
-		s.lidsTable.StartBlockIndex = blocksCounter
-		for block := range bb.BuildLIDsBlocks(src.TokenLIDs(), consts.LIDBlockCap) {
-			if !push(s.packLIDsBlock(block), &statsLIDs) {
-				return
-			}
-		}
-		if s.lastErr = util.CollapseErrors([]error{src.LastError(), bb.LastError()}); s.lastErr != nil {
-			return
-		}
+func (s *IndexSealer) WriteInfoFile(ws io.WriteSeeker, src Source) error {
```
eguguchkin: I already mentioned that the info block here is always a single block and always uncompressed.

In this case, we don't need headers, nor the registry and its cache, since info is read once and remembered forever.

What if the *.info file were always plain JSON?

Future considerations:
 - this would help us get rid of the fraction_info_cache file
 - and if we store info for unloaded fractions as well, it would solve the problem of the workaround in the loader: `hasIndex, err := l.provider.s3cli.Exists(ctx, indexName)`

frac/index_cache.go
```diff
@@ -8,7 +8,14 @@ import (
...
+	TokenRegistry   *cache.Cache[[]byte]
+	OffsetsRegistry *cache.Cache[[]byte]
```
eguguchkin, line 35: Offsets are a single block, and like info, we load it only once and keep it all the time. I suggest not using either a reader or a cache for the registry.

frac/active_sealing_source.go
```diff
@@ -24,72 +23,68 @@ import (
...
+	docsReader   *storage.DocsReader   // Document storage reader
 }
 
-// NewActiveSealingSource creates a new data source for sealing
-// based on an active in-memory index.
 func NewActiveSealingSource(active *Active, params common.SealParams) (*ActiveSealingSource, error) {
 	info := *active.info // copy
+
 	sortedLIDs := active.GetAllDocuments()
+	fields, fieldTids := sortFields(active.TokenList)
```
eguguchkin: fieldTids => fieldTIDs for consistency in naming across the entire project.

frac/active_sealing_source.go
```diff
@@ -24,72 +23,68 @@ import (
...
+	blocksOffsets []uint64 // Document block offsets
+
+	sortedLIDs   []uint32 // Sorted LIDs (Local ID)
+	oldToNewLIDs []uint32 // Mapping from old LIDs to new ones (after sorting)
+
+	mids *UInt64s // MIDs
+	rids *UInt64s // RIDs
+
+	fields    []string   // Sorted field names
+	fieldTids [][]uint32 // Each field contains sorted TIDs based on token value
```
eguguchkin: fieldTids => fieldTIDs

frac/active_sealing_source.go
```diff
@@ -101,49 +96,69 @@ func NewActiveSealingSource(active *Active, params common.SealParams) (*ActiveSe
...
 
-// sortFields sorts field names and calculates maximum TIDs for each field.
-// Returns sorted field list and array of maximum TIDs.
-func sortFields(tl *TokenList) ([]string, []uint32) {
-	fields := make([]string, 0, len(tl.FieldTIDs))
-	for field := range tl.FieldTIDs {
-		fields = append(fields, field)
-	}
+func sortFields(tl *TokenList) ([]string, [][]uint32) {
+	fields := slices.Collect(maps.Keys(tl.FieldTIDs))
```
eguguchkin, line 100: nit: This form looks concise and clear, but I have some doubts:

- First, in the Collect implementation, the slice isn't pre-allocated to the needed size, so we lose performance on grow.
- Second, Keys works via an iterator, and we pay some cost for that.

Yes, the number of fields is unlikely to exceed a few thousand right now, but this wouldn't be a great implementation for a hot path. What do you think?

frac/active_sealing_source.go
```diff
@@ -155,130 +170,47 @@ func (src *ActiveSealingSource) prepareInfo() {
 	src.info.BuildDistribution(mids)
 }
 
-// Info returns index metadata information.
-func (src *ActiveSealingSource) Info() *common.Info {
-	return src.info
+func (src *ActiveSealingSource) prepareLids() {
```
eguguchkin, line 173: nit: Just discussing here — what effect is expected from this rearrangement of GetLIDs calls?

frac/sealed/sealing/sealer.go
```diff
@@ -13,88 +13,156 @@ import (
...
-	Fields() iter.Seq2[string, uint32]                    // Ordered sequence of fields with their max field's TID value
-	TokenLIDs() iter.Seq[[]uint32]                        // Sequence of Token LIDs ordered by TID and LID
-	BlocksOffsets() []uint64                              // Offsets of DocBlock's in the doc file
-	LastError() error                                     // Last error encountered during data retrieval
+	// Info returns metadata describing this source.
+	Info() *common.Info
+
+	// ID returns an iterator over stored document identifiers paired with
+	// their positions, in descending [seq.ID] order.
+	ID() iter.Seq2[DocLocation, error]
```
eguguchkin, line 29: ID => IDs ?

frac/sealed.go
```diff
@@ -39,9 +37,25 @@ type Sealed struct {
...
-	indexFile   *os.File
-	indexCache  *IndexCache
-	indexReader storage.IndexReader
+	// IsLegacy is true for fractions that use the old single .index file format.
+	IsLegacy     bool
+	legacyFile   *os.File
+	legacyReader storage.IndexReader
+
+	// Per-section index files and their readers (new split format only).
+	infoFile    *os.File
```
eguguchkin, line 47: It seems we don't need to store infoFile, offsetFile, infoReader, offsetReader directly as struct fields? It's enough to read the files once when loading the fraction, load the contents, and close the files.

We constantly reuse only three readers in your implementation (see `createDataProvider`):

```go
tokenReader := &f.tokenReader
lidReader := &f.lidReader
idReader := &f.idReader
```

These are the ones we should store as fields of sealed.

frac/sealed.go
```diff
@@ -88,44 +104,140 @@ func NewSealed(
...
+				zap.String("file", name),
+				zap.Error(err),
+			)
 		}
-		f.indexReader = storage.NewIndexReader(f.readLimiter, f.indexFile.Name(), f.indexFile, f.indexCache.Registry)
+
+		f.legacyFile = file
+		f.legacyReader = storage.NewIndexReader(
+			f.readLimiter, file.Name(),
+			file, f.indexCache.InfoRegistry,
```
eguguchkin: So we didn't create a separate LegacyRegistry but reused another one that was intended for Info? Instead, it now looks like a typo and confuses the reader. Is it worth it?

frac/sealed.go
```diff
@@ -88,44 +104,140 @@ func NewSealed(
...
 
-	// fast path if fraction-info cache exists AND it has valid index size
+	// Fast path: if info cache has valid index size, skip opening the info file now.
 	if info != nil && info.IndexOnDisk > 0 {
 		return f
 	}
 
-	f.openIndex()
-	f.info = loadHeader(f.indexFile, f.indexReader)
+	f.openInfo()
```
eguguchkin: Let's explicitly create two branches here and elsewhere for legacy and splitted - otherwise the logic becomes very non-obvious with implicit substitutions (// infoReader is used by [loadInfo]), and that's a cognitive trap.

I'm suggesting something like:

```diff
diff --git a/frac/sealed.go b/frac/sealed.go
index c7c92023..42fbba84 100644
--- a/frac/sealed.go
+++ b/frac/sealed.go
@@ -109,15 +109,21 @@ func NewSealed(
                return f
        }

+       if isLegacy {
+               f.openLegacyIndex()
+               f.info = loadInfo(f.legacyReader)
+
+       } else {
                f.openInfo()
                f.info = loadInfo(f.infoReader)
+       }
+
        f.info.IndexOnDisk = computeIndexOnDisk(f.BaseFileName, f.IsLegacy)

        return f
 }

-func (f *Sealed) openInfo() {
-       if f.IsLegacy {
+func (f *Sealed) openLegacyIndex() {
        if f.legacyFile != nil {
                return
        }
@@ -143,6 +149,7 @@ func (f *Sealed) openInfo() {
        return
 }

+func (f *Sealed) openInfo() {
        if f.infoFile != nil {
                return
        }
@@ -165,11 +172,13 @@ func (f *Sealed) openInfo() {
 }

 func (f *Sealed) openIndex() {
-       f.openInfo()
        if f.IsLegacy {
+               f.openLegacyIndex()
                return
        }

+       f.openInfo()
+
        if f.tokenFile == nil {
                name := f.BaseFileName + consts.TokenFileSuffix
                file, err := os.Open(name)

```

frac/sealed.go
```diff
@@ -180,18 +292,31 @@ func (f *Sealed) load() {
...
+		return
+	}
 
-		f.openDocs()
-		f.openIndex()
+	f.openDocs()
+	f.openIndex()
 
-		(&Loader{}).Load(&f.blocksData, f.info, &f.indexReader)
+	if f.IsLegacy {
```
eguguchkin, line 303: I think explicit separation is required here as well.



```go
func (f *Sealed) load() {
	f.loadMu.Lock()
	defer f.loadMu.Unlock()

	if f.isLoaded {
		return
	}

	f.openDocs()
	
	if f.IsLegacy {
		f.openLegacyIndex()
		(&LegacyLoader{}).Load(&f.blocksData, f.info, f.legacyReader)
		f.isLoaded = true
		return
	}
	
	f.openIndex()
	(&Loader{}).Load(&f.blocksData, f.info, IndexReaders{
		Info:    f.infoReader,
		Token:   f.tokenReader,
		Offsets: f.offsetsReader,
		ID:      f.idReader,
		LID:     f.lidReader,
	})

	f.isLoaded = true
}
```

frac/sealed.go
```diff
@@ -370,39 +537,48 @@ func (f *Sealed) IsIntersecting(from, to seq.MID) bool {
...
 	}
 
 	var bi sealed.BlockInfo
 	if err := bi.Unpack(block); err != nil {
-		logger.Fatal(
-			"error unpacking info block",
-			zap.String("file", indexFile.Name()),
-			zap.Error(err),
-		)
+		logger.Fatal("error unpacking info block", zap.Error(err))
```
eguguchkin: Here we'll definitely be missing information about which file caused the error, and we need that information to later investigate the problem on a specific file.

And probably above as well.

frac/remote.go
```diff
@@ -231,34 +267,133 @@ func (f *Remote) load() error {
...
+		ID:      f.idReader,
+		LID:     f.lidReader,
+	})
+
+	f.isLoaded = true
 	return nil
 }
 
+func (f *Remote) openInfo() error {
+	if f.IsLegacy {
```
eguguchkin: I have the same suggestions here as for sealed:

- explicitly separate opening of new and old indexes into different methods
- don't reuse infoRegistry as a cache for legacy
- explicitly open either the old index or the new index everywhere

frac/sealed_loader.go
```diff
@@ -56,104 +61,231 @@ func (l *Loader) Load(blocksData *sealed.BlocksData, info *common.Info, indexRea
...
+		l.blockIndex += 2 // skip RIDs and Pos blocks
 	}
 
-	return idsTable, blocksOffsets, nil
+	return table, offsets.Offsets, nil
 }
 
-func (l *Loader) skipTokens() {
+// loadLIDs scans LID block headers, recording the absolute start index for lids.Table.
+func (l *LegacyLoader) loadLIDs() (*lids.Table, error) {
```
eguguchkin: nit: There's an inconsistency here — functions that are identical in meaning and return type are named differently in the new loader and the old one.

Compare:

`func (l *Loader) loadLIDsTable(r storage.IndexReader) (*lids.Table, error) {`

frac/sealed_loader.go
```diff
@@ -56,104 +61,231 @@ func (l *Loader) Load(blocksData *sealed.BlocksData, info *common.Info, indexRea
...
+
+// loadIDsTable scans block headers in the .id file to build seqids.Table.
+// Blocks are stored as (MIDs, RIDs, Pos) triplets; we only need MIDs headers.
+func (l *Loader) loadIDsTable(r storage.IndexReader, idsTotal uint32, fracVersion config.BinaryDataVersion) seqids.Table {
+	table := seqids.Table{
+		StartBlockIndex: 0,
+		IDsTotal:        idsTotal,
+	}
+
+	for blockIdx := uint32(0); ; {
```
eguguchkin: The empty block at the end was left only for compatibility with the old index file reading code. But over time, old indexes will disappear, this empty marker will become meaningless, and there will be no motivation to change the format again (to remove the unnecessary empty block). I suggest we get rid of the unnecessary empty block right now and rewrite the reading code as follows:

```go
for blockIdx := uint32(0); blockIdx < r.TotalBlocks(); blockIdx += 3 {
```

frac/sealed_loader.go
```diff
@@ -56,104 +61,231 @@ func (l *Loader) Load(blocksData *sealed.BlocksData, info *common.Info, indexRea
...
-	for {
-		// skip token table
-		header := l.skipBlock()
-		if header.Len() == 0 {
+	return lids.NewTable(startIndex, minTIDs, maxTIDs, isContinued), nil
+}
+
+// IndexReaders holds one IndexReader per split index file.
+type IndexReaders struct {
+	Info    storage.IndexReader
```
eguguchkin: Unused

fracmanager/cache_maintainer.go
```diff
@@ -149,7 +149,12 @@ func (cm *CacheMaintainer) CreateIndexCache() *frac.IndexCache {
 		LIDs:       newCache[*lids.Block](cm, lidsName),
 		Tokens:     newCache[*token.Block](cm, tokensName),
 		TokenTable: newCache[token.Table](cm, tokenTableName),
-		Registry:   newCache[[]byte](cm, indexName),
+		// Each index file gets its own registry cache (they all use key=1 internally).
+		InfoRegistry:    newCache[[]byte](cm, indexName),
```
eguguchkin: InfoRegistry, OffsetsRegistry - not needed here

frac/sealed/sealing/blocks_builder.go
dkharms: Yep, I've removed it in #401

frac/sealed/sealing/blocks_builder.go
dkharms: What such change would make better? In my opinion, it will just make this function more complex since we would need build tokens blocks and lid blocks at the same time in the **same** function.

dkharms: Not sure about this one as well. Now all state is scoped to the specific functions and this state is not shareable -- I would like to leave it the way it is now.

dkharms: This comment was a duplicate -- https://github.com/ozontech/seq-db/blob/main/frac/sealed/token/table.go

fracmanager/loader.go
dkharms: > By the way, it makes sense to bump the fraction version stored in info, and then determine whether it's legacy or not from info itself.

Nope, that would not help. In order to determine fraction binary version we have to already understand from which file this information will be read -- `.index` or `.info`. So basically, what I am doing right now.

dkharms: > This looks like a workaround. Do you have an understanding of how to solve this properly in the future?

To solve what?

Maybe, we should store content of `.info` inside `.remote` file? Otherwise on startup we generate a lot of requests to S3 just to understand boundaries (`from`, `to`) of fractions.

And the more fractions we have, the more requests we have to perform. Also, in this case we could just bump fraction binary version.

frac/sealed/sealing/writer.go
dkharms: Yep, I'll go with `blockTypeInfo`.

frac/sealed/sealing/writer.go
dkharms: > But overall, having an empty block at the end is questionable for a sharded index.

Not empty block -- it is empty header block. How else would you simulate EOF?

I agree that this is not the cleanest way but all current code base is already works with this assumption so it was just much easier to leave it this way.

dkharms: > How about extracting this into a separate method?

Yep, will do that.

frac/sealed/sealing/sealer.go
dkharms: Then I would change `TokenTriplet` to `TokenTriplets` as well, right?

frac/sealed_loader.go
dkharms: Yep, will rename it.

seq/seq.go
dkharms: Well, such variables were used across different packages (`frac`, `sealing`) and duplicated each other.

frac/sealed/sealing/blocks_builder.go
dkharms: Fixed in #414

frac/sealed/sealing/blocks_builder.go
dkharms: Fixed in #414

frac/sealed/sealing/index.go
dkharms: The part with `lidAccumulator` is fixed in #414.
Other issues (specifically, empty blocks at the end of files) will be addressed in other PR.

frac/sealed/sealing/index.go
dkharms: Yep, this was a typo. Fixed in #414.

frac/sealed/sealing/sealer.go
dkharms: I would stick with singular form since iterators return one ID at a time.

frac/sealed/sealing/writer.go
dkharms: Fixed in #414.

frac/sealed/sealing/writer.go
dkharms: Fixed in #414.

frac/sealed_loader.go
dkharms: Fixed in #414.

util/pair.go
dkharms: Decided to stick with the current implementation. There is a one more PR is incoming (#401) where type aliases heavily exploited -- if you are not OK with this we can come up with better solution in that (or following) PRs.

frac/active_sealing_source.go
dkharms: Fixed in #414.

frac/active_sealing_source.go
dkharms: Fixed in #414.

frac/sealed/sealing/blocks_builder.go
dkharms: Fixed in #414

eguguchkin (approved):

PR #390, opened 2026-04-01
feat: batching execution (single iterator queries)

cheb0:
# Description

No significant perf improvement, just introduce new types: `LIDBatch`, `BatchedNode`. Batch mode enabled for single-iterator queries for sealed fractions.

`iterateEvalTree` is now works with batches. For one-by-one lid we batch them a bit, for batched flow we just bypass batches as is. To batch one-by-one eval tree lids, a buffer is used. It preallocated with 64k slots.

`LIDBatch` currently stores lids as `[]uint32` sorted asc. I had to introduce an unneeded copying from LID batch format to slice of `node.LID` so that we have a single iterateEvalTree function. It's a cost of avoiding code duplication.

### single-iterator query perf

For single-iterator query we transfer batches directly. Hot query perf `service:X` (hist, total is 1.6 million):

main branch:
**29.578 ms**

PR:
**23.854 ms**

### stopwatch eval_tree_next fixed

As part of the PR, stopwatch usage for `eval_tree_next` is adrressed. For low-cardinality query `service:X AND level:4` (histogram) it reports the following numbers:

Search (hot) completed. Found 100 documents. Took: 15.812 ms
total >> iterate_eval_tree >> eval_tree_next -> **95.407644ms**

With this PR:
Search (hot) completed. Found 100 documents. Took: 13.958 ms
total >> iterate_eval_tree >> eval_tree_next -> **5.5113ms**

Part of #329 

---


cheb0:

cheb0:
@seqbenchbot down 73ed4722

frac/processor/search.go
```diff
@@ -59,6 +75,10 @@ func IndexSearch(
 			return evalLeaf(index, token, sw, stats, minLID, maxLID, params.Order)
 		},
 	)
+	if err != nil {
+		return nil, err
+	}
```
forshev: seem like we already have error check right after m.Stop() call

frac/processor/search.go
```diff
@@ -107,8 +129,30 @@ func IndexSearch(
 		m.Stop()
 	}
 
+	var evalTreeIter func(need int, out lidsBuf) LIDsIter
+	batchNode, ok := tryConvertToBatchedTree(evalTree)
+
+	if ok {
+		evalTreeIter = func(need int, _ lidsBuf) LIDsIter {
```
forshev: nit: `need` is unused and can be replaced with `_`

forshev (approved):

eguguchkin (approved):

frac/processor/search.go

PR #395, opened 2026-04-02
chore: copy corrupted WAL and docs to .broken dir

cheb0:
# Description

If WAL file is corrupted, then increment a metric and copy to a separate directory `.broken`

---


util/fs.go
```diff
@@ -53,3 +54,24 @@ func RemoveFile(file string) {
...
+
+// CopyFile copies a file (overwrites if already exists)
+func CopyFile(src, dst string) error {
+	in, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer in.Close()
+
+	out, err := os.OpenFile(dst, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0o666)
```
dkharms: Why write-only? I mean this is not a good default for a such generic function `CopyFile`.

util/fs.go
```diff
@@ -53,3 +54,24 @@ func RemoveFile(file string) {
...
+	if err != nil {
+		return err
+	}
+
```
dkharms, line 160: `defer out.Close()` instead of manually closing the file?

dkharms (approved):

moflotas (approved):

PR #397, opened 2026-04-03
perf: improve wildcard query perf with predicate and contains-check pushdown 

cheb0:
# Description

Currently we spend only a fraction of time calling `bytes.Index`. This PR partially addresses that.

This PR pushes `pattern.Searcher` to `Block` level, so that `Block` is able to stream tokens through searcher. For ordinary wildcards like `*error*` there is direct `FindContains` method which is even faster.

For example, query `message:*foobarf*`:
main: 86 ms
using `FindToken`: 50 ms
using `FindContains`: 37 ms

So, FindContains just throws out costly abstractions to get additional performance. We could also provide a dedicated func like FindSuffix, for example.  This is a typical example when performance requires additional code.

<!DOCTYPE html>
<html>
<body>

Query | Type | Ids | cold, ms | hot, ms | cold (branch), ms | hot (branch), ms | cold diff | hot diff
-- | -- | -- | -- | -- | -- | -- | -- | --
`trace_id:*foobar` | reg | 0 | 18.76 | 4.37 | 16.14 | 1.84 | -14% | -57.9%
`k8s_pod:*6` | reg | 100 | 13.3 | 0.67 | 13.03 | 0.47 | -2% | -29.9%
`message:*err*` | reg | 100 | 138.72 | 26.97 | 120.27 | 12.36 | -13.3% | -54.2%
`message:*foo*` | reg | 100 | 77.69 | 27.08 | 60.54 | 11.84 | -22.1% | -56.3%
`message:*request*` | reg | 100 | 124.95 | 25.45 | 104.13 | 10.37 | -16.7% | -59.3%
`message:*foobar*foobar*` | reg | 0 | 187.54 | 64.25 | 147.31 | 30.5 | -21.5% | -52.5%
`message:*foobarfoobar*` | reg | 0 | 184.93 | 63.87 | 121.51 | 20.39 | -34.3% | -68.1%
`message:*very_very_message_aggregator_events*` | reg | 0 | 173.45 | 51.62 | 116.9 | 12.81 | -32.6% | -75.2%
</body>
</html>

Next steps:
- try calling `bytes.Index` over `Block` payload - already shows good results
- build Offsets lazy - if previous is done
- modernize token Block, boost Unpack speed

---


frac/sealed/token/block_loader.go
```diff
@@ -60,6 +62,30 @@ func (b *Block) GetToken(index int) []byte {
 	return b.Payload[offset : offset+l]
 }
 
+func (b *Block) FindContains(from, to int, needle []byte) ([]int, error) {
```
dkharms: We've discussed that you can perform `bytes.Contains` on the block payload before checking each token individually. Have you measured performance of such optimization?

frac/sealed/token/block_loader.go
```diff
@@ -60,6 +62,30 @@ func (b *Block) GetToken(index int) []byte {
...
+	indices := make([]int, 0)
```
dkharms: I guess you could pass here slice of needles as well to handle queries like `message:*foo*bar*` with multiple needles. Or there is something that blocks such improvement?

frac/sealed/token/block_loader.go
cheb0: > We've discussed that you can perform bytes.Contains on the block payload before checking each token individually.

Yes, I tried calling `bytes.Index` on entire payload. It boosts even further comparing to this PR:
message:*foobar* 
35 ms => 9 ms

However, this means that when `bytes.Index` returns and if we have some proper index returned, then we need to do a bin search on `Offsets` to find an index and then check for false positive. It also comes with neat property that we can avoid call `Unpack` (build offsets) lazily which boosts cold query performance (somewhat around extra 20%).

I put a task to the backlog, decided that it's too much for a single PR.

frac/sealed/token/block_loader.go
cheb0: No, I think it's doable. Maybe will do

cheb0: upd: will do in a separate PR

frac/sealed/token/block_loader.go
dkharms: I guess it's better to rename `Block.FindContains` and `Block.FindToken` for several reasons:
- The name is conflicting with what we have in `tokenProvider` interface however they serve different purpose;
- These methods should be private, in my opinion;

Maybe something like will be better?
```go
func (b *Block) contains(from, to int, needle []byte) ([]int, error) { ... }
func (b *Block) find(from, to int, searcher pattern.Searcher) ([]int, error) { ... }
```

pattern/pattern.go
```diff
@@ -17,6 +17,8 @@ import (
 
 type tokenProvider interface {
 	GetToken(uint32) []byte
+	FindContains(firstTID uint32, lastTID uint32, needle []byte) ([]uint32, error)
```
dkharms: Why did you decide to make `firstTID` and `lastTID` a part of an API?

Seems like for this specific case (e.g. query `foo:'*bar*'`) we cannot narrow the TID search boundaries.

And now we always pass the first and last TID in this method:
https://github.com/ozontech/seq-db/blob/0-wildcard-predicate-pushdown/pattern/pattern.go#L411

frac/sealed/token/provider.go
```diff
@@ -55,3 +57,71 @@ func (tp *Provider) GetToken(tid uint32) []byte {
...
+	lastIdx := sort.Search(len(tp.entries), func(i int) bool {
+		return tp.entries[i].StartTID > lastTID
+	})
+	lastIdx--
+	if lastIdx < firstIdx {
+		return nil
+	}
+	entries := tp.entries[firstIdx : lastIdx+1]
+	return entries
+}
```
dkharms: It is totally safe to rewrite this method in this way and it raises less questions on why we perform decrement and increment in `lastIdx`:
```go
func (tp *Provider) narrowEntries(firstTID, lastTID uint32) []*TableEntry {
	firstIdx := sort.Search(len(tp.entries), func(i int) bool {
		return tp.entries[i].getLastTID() >= firstTID
	})

	if firstIdx >= len(tp.entries) {
		return nil
	}

	lastIdx := sort.Search(len(tp.entries), func(i int) bool {
		return tp.entries[i].StartTID > lastTID
	})

	// INVARIANT: Following condition always holds:
	// lastIdx <= len(tp.entries) && firstIdx <= lastIdx
	return tp.entries[firstIdx:lastIdx]
}
```

frac/sealed/token/provider.go
```diff
@@ -55,3 +57,71 @@ func (tp *Provider) GetToken(tid uint32) []byte {
...
+	entries := tp.narrowEntries(firstTID, lastTID)
+	if len(entries) == 0 {
+		return nil, nil
+	}
+
+	var tids []uint32
+
+	for _, entry := range entries {
+		block := tp.findBlock(entry.BlockIndex)
+		firstIndex, lastIndex := tp.narrowTIDs(entry, firstTID, lastTID)
```
dkharms: Seems like it is beneficial to narrow tids only for the first and last entries -- for everything in-between it is just an additional overhead on method call.

And I guess this is name is ambiguous as well -- what we really do here is deriving local index of token inside specific block from its universal tid.

frac/sealed/token/provider.go
```diff
@@ -55,3 +57,71 @@ func (tp *Provider) GetToken(tid uint32) []byte {
...
+		}
+		for _, idx := range indices {
+			tid := entry.StartTID + uint32(idx-int(entry.StartIndex))
+			tids = append(tids, tid)
+		}
+	}
+	return tids, nil
+}
+
+func (tp *Provider) narrowTIDs(entry *TableEntry, firstTID, fromTID uint32) (int, int) {
```
dkharms: Incorrect argument name `fromTID` -- I guess you've meant `lastTID` here.
And I suggest to use builtin functions for getting min/max:
```go
func (tp *Provider) narrowTIDs(entry *TableEntry, firstTID, lastTID uint32) (int, int) {
	tidStart := max(firstTID, entry.StartTID)
	tidEnd := min(lastTID, entry.getLastTID())

	firstIndex := entry.GetIndexInTokensBlock(tidStart)
	lastIndex := entry.GetIndexInTokensBlock(tidEnd)

	return firstIndex, lastIndex
}
```

pattern/pattern.go
```diff
@@ -390,22 +392,24 @@ func isNaNOrInf(f float64) bool {
...
-	for tid := s.firstTID(); tid <= s.lastTID(); tid++ {
-		if tid&1023 == 0 && util.IsCancelled(ctx) {
-			return nil, ctx.Err()
-		}
-
-		match, err := s.check(tp.GetToken(tid))
-		if err != nil {
-			return nil, err
-		}
+func isSimpleWildcardContains(token parser.Token) (needle []byte, ok bool) {
```
eguguchkin, line 396: Please add a comment with an example of an expression that fits/matches this check

frac/sealed/token/provider.go
```diff
@@ -55,3 +57,71 @@ func (tp *Provider) GetToken(tid uint32) []byte {
...
+	if len(entries) == 0 {
+		return nil, nil
+	}
+
+	var tids []uint32
+
+	for _, entry := range entries {
+		block := tp.findBlock(entry.BlockIndex)
+		firstIndex, lastIndex := tp.narrowTIDs(entry, firstTID, lastTID)
+		indices, err := search(block, firstIndex, lastIndex)
```
eguguchkin: nit: better to use 'indexes' in this context


Option | Context of use | Examples
-- | -- | --
indexes | Recommended variant for databases, books, and general context. | database indexes, book indexes, market indexes
indices | Preferred in mathematics, science, and engineering. | array indices, price indices, mathematical indices

frac/sealed/token/provider.go
eguguchkin: This is not a token provider method; tp is not used in this function at all. Rather, this is a method for TableEntry

eguguchkin (approved):

frac/sealed/token/block_loader.go
cheb0: renamed

pattern/pattern.go
cheb0: For `foo:'*bar*'` we narrow down to all tokens of `foo` field, i.e. `foo:?` tokens. If there are 50 such tokens only, we will check only a single block, and `firstTID` might be like 1000 and `lastTID` 1050

frac/sealed/token/provider.go
cheb0: copied. hope it works

frac/sealed/token/provider.go
cheb0: It's now `narrowIndexes`. Regarding the performance cost - it's called once per block only. After all other optimizations I check if there is more to squeeze out of it.

frac/sealed/token/provider.go
cheb0: fixed

cheb0: moved to `TableEntry`

pattern/pattern.go
dkharms: Yeah, I see. But for specific case of `FindContains()` arguments `firstLID`, `lastLID` should not be a part of an API. Here, take a look, we already have all necessary information in `Provider`:
```go
type Provider struct {
	...
    // NOTE: Entries already were narrowed.
	entries  []*TableEntry
	...
}

func (tp *Provider) FirstTID() uint32 {
	return tp.entries[0].StartTID
}

func (tp *Provider) LastTID() uint32 {
	return tp.entries[len(tp.entries)-1].getLastTID()
}

func (tp *Provider) FindContains(needle []byte) ([]uint32, error) {
	return tp.findInBlocks(tp.FirstTID(), tp.LastTID(), func(b *Block, firstIndex, lastIndex int) ([]int, error) {
		return b.contains(firstIndex, lastIndex, needle)
	})
}
```

dkharms (approved):

PR #405, opened 2026-04-17
feat: implement LID block size config

### LINKED ISSUE #330: Configurable block sizes in sealed fraction
For search requests on real seq-db instances there are two major contributors of CPU usage:
- iterating through query tree (left side on the flamegraph)
- reading LID blocks from disk (right side on the flamegraph)

Image

The first part will be addressed through batcher query execution and block skipping. The second part can be partially addressed though lower LID pages. Databases strive to align page size to 4kb which is equal to the minimum amount which can be read from disk.

Pros:
- higher cache granularity and efficiency
- better skipping granurality
- faster performance on search query and histograms

Cons:
- more independent disk reads on aggregations
- more independent read on S3 searches

Most cons can be addressed through intelligent prefetching and repacking fracs before uploading to S3.

cheb0:
# Description

Allow to configure LID block size.

Fixes #330

---


config/config.go
```diff
@@ -200,6 +200,8 @@ type Config struct {
 		MetasZstdCompressionLevel    int `config:"metas_zstd_compression_level" default:"1"`
 		SealedZstdCompressionLevel   int `config:"sealed_zstd_compression_level" default:"3"`
 		DocBlockZstdCompressionLevel int `config:"doc_block_zstd_compression_level" default:"3"`
+		// LIDBlockSize sets max lids (postings) saved per LIDs block.
+		LIDBlockSize Bytes `config:"lid_block_size" default:"64KiB"`
```
dkharms: I suggest to get rid of this nonsense with `Bytes` because we do not build blocks by examining size of block but just by examining count of LIDs inside a block:
```go
for _, lid := range lidsbuf {
	// Flush if there is more LIDs than specified threshold.
	if len(a.currentBlock.payload.LIDs) == a.blockCapacity {
		if err := a.onBlock(a.finalizeBlock()); err != nil {
			return err
		}

		a.currentBlock.payload.LIDs = a.currentBlock.payload.LIDs[:0]
		...
	}

	...
}
```

Migration is gonna be painless since this is how we store this information in `.info` file:
```
seq-db ⟩ cat /tmp/tmp.mCJYmApMn2/seq-db-01KREF7BBEPZ0KB0HZZ1CCKP33.info
SEQM{..., "const_lid_block_cap":65536, ...}
``` 

What do you think? This is bugging me for a very long time, honestly.

dkharms (approved):

cmd/seq-db/seq-db.go
```diff
@@ -283,6 +284,7 @@ func startStore(
 				},
 				SkipSortDocs: !cfg.DocsSorting.Enabled,
 				KeepMetaFile: false,
+				LIDBlockSize: int(cfg.Compression.LIDBlockSize),
```
eguguchkin: Let's remove this field: `Fraction.LIDBlockSize`. It is used in a chain that has no functional value: this parameter initializes `Info.ConstLIDBlockCap`, which is **ultimately not used anywhere** in the code. Its potential use is for manual debugging: from the console – read the contents of the info block on disk: `cat <frac_name>.info`.

I propose not to initialize this field at all for the active faction (=0). Initialize it during sealing (see `ActiveSealingSource.prepareInfo()` method). In the info structure, add a comment that the field is only filled when writing to disk and is not used in the code (so that someone in the future does not build logic around the empty field of the active faction).

config/config.go
eguguchkin: nit: Why in the compression section? It seems like we need to make a separate section like Sealing or InvertedIndex or Fraction. And move Compression there as well.

eguguchkin (approved):

config/config.go
cheb0: made `Sealing` section. So config is now sealing.lids.block_size. Compression might be moved there in future as well

cmd/seq-db/seq-db.go
cheb0: I though about the same. Removed

config/config.go
cheb0: done

PR #407, opened 2026-04-24
perf: skip lid blocks

cheb0:
# Description

### Summary

- new frac version: `v5`
- LID block ext1 now has `firstLID` and `lastLID` encoded
- `isContinued` is only used for older frac versions
- no legacy format TID adjusting for a new frac version
- as previously, we always locate and load the first block eagerly but calling `NextGeq` allows to skip some blocks

### Measurements

With 16k LID blocks, there are only 4 search requests (out of 50+ I tested) where this feature improves cold query perf. All four search requests fit into "intersect dense and sparse column" type. 

<!DOCTYPE html>

<html>
<body>

Query | env | Total | cold, ms |   | LID blocks read | cold (branch), ms |   | LID blocks read (branch) | cold diff | LID blocks read, diff | Trial_1 | Trial_2 | Trial_3
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
service:E and login:L and moduleName:S and (zone:A or zone:B or zone:C) and _exists_:message | prod | 0 | 22.29 | ±1.03 | 164 | 20.6 | ±1.10 | 62 | -7.6% | -62.2% | -12.6% | -8.9% | -7.6%
request_host:huge AND remote_addr:X (found) | lb | 0 | 18.22 | ±0.97 | 69 | 17.53 | ±0.82 | 42 | -3.8% | -39.1% | -3.8% | -9.8% | -3.8%
request_host:small AND response_status:200 | lb | 69 | 33.64 | ±1.92 | 248 | 30.44 | ±1.60 | 138 | -9.5% | -44.4% | -2% | -10.9% | -9.5%
response_status:500 \| by request_method | lb | 172 | 52.58 | ±2.14 | 377 | 51.68 | ±1.54 | 333 | -1.7% | -11.7% | -7.9% | -20.6% | -1.7%

</body>
</html>

---

### DIFF
--- a/frac/sealed/lids/table.go
+++ b/frac/sealed/lids/table.go
@@ -98,9 +98,9 @@ func (t *Table) SeekBlockGeq(index, tid, nextLID uint32) uint32 {
 	}
 
 	res := index
-	for i := index + 1; i < uint32(len(t.MinTIDs)); i++ {
+	for i := int(index) + 1; i < len(t.MinTIDs); i++ {
 		if t.MinTIDs[i] == tid && nextLID >= t.FirstLIDs[i] {
-			res = i
+			res = uint32(i)
 			continue
 		}
 		break
--- a/frac/sealed/lids/table.go
+++ b/frac/sealed/lids/table.go
@@ -39,7 +39,7 @@ func NewTable(
 }
 
 func (t *Table) GetAdjustedMinTID(blockIndex uint32) uint32 {
-	if t.FracVer < config.BinaryDataV5 {
+	if t.FracVer < config.BinaryDataV6 {
 		if t.IsContinued[blockIndex] {
 			return t.MinTIDs[blockIndex] - 1
 		}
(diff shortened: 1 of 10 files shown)


frac/sealed/lids/table.go
```diff
@@ -75,6 +87,46 @@ func (t *Table) GetLastBlockIndexForTID(tid uint32) uint32 {
...
+func (t *Table) SeekBlockGeq(index uint32, tid uint32, nextLID uint32) uint32 {
+	if t.FracVer < config.BinaryDataV4 {
+		// not supported for old frac versions
+		return index
+	}
+
+	res := index
+	for i := index + 1; i < uint32(len(t.MinTIDs)); i++ {
+		if t.MinTIDs[i] == tid && nextLID >= t.FirstLIDs[i] {
+			res = i
```
forshev: nit: here `i` is int, but in `SeekBlockLeq()` it is uint32. let's make it consistent, for example:

```go
for i := int(index) + 1; i <len(t.MinTIDs); i++ {
	if t.MinTIDs[i] == tid && nextLID >= t.FirstLIDs[i] {
		res = uint32(i)
		continue
	}
	break
}
```

forshev (approved):

eguguchkin (approved):

eguguchkin (approved):

frac/sealed/lids/table.go
cheb0: fixed

PR #410, opened 2026-04-27
refactor: do not store ids count in offsets

dkharms:
# Description                                                                                   
                                                                                                                               
Previously, `BlockOffsets` stored the total number of document IDs alongside the block offsets. This meant `WriteOffsetsFile` had to read from `Info`, which is a problem for compaction — we don't have complete `Info` while building the index.                

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/frac/sealed_loader.go
+++ b/frac/sealed_loader.go
@@ -36,7 +36,7 @@ func (l *LegacyLoader) Load(blocksData *sealed.BlocksData, info *common.Info, re
 	l.skipSection() // skip token table blocks
 
 	var err error
-	blocksData.IDsTable, blocksData.BlocksOffsets, err = l.loadIDs(info.BinaryDataVer)
+	blocksData.IDsTable, blocksData.BlocksOffsets, err = l.loadIDs(info)
 	if err != nil {
 		logger.Fatal("legacy load ids error", zap.Error(err))
 	}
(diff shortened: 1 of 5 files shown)


cheb0 (approved):

forshev (approved):

frac/sealed_loader.go
```diff
@@ -94,9 +94,8 @@ func (l *LegacyLoader) loadIDs(fracVersion config.BinaryDataVersion) (seqids.Tab
 	l.blockIndex++
 
 	table := seqids.Table{
-		StartBlockIndex: l.blockIndex, // absolute index of first MID block in .index
-		IDsTotal:        offsets.IDsTotal,
-		IDBlocksTotal:   uint32(len(offsets.Offsets)),
+		StartBlockIndex: l.blockIndex,       // absolute index of first MID block in .index
+		IDsTotal:        info.DocsTotal + 1, // Increment by one for [seq.SystemID]
```
eguguchkin, line 98: Please add comments at the declaration sites of `seqids.Table.IDsTotal` and `Info.DocsTotal` to clarify that the stated value may exceed the actual number of unique documents in the fraction. 

This is because, when a nested index is used, a single document with one seq.ID can be associated with multiple LIDs. Without such clarification, one might intuitively assume that `DocsTotal` represents the exact document count, but that is not the case.

This nuance is non‑obvious, and we (I) keep forgetting about it, which makes me concerned that we might inadvertently introduce bugs in the future if we overlook the implications of nested indexes.

eguguchkin (approved):

frac/sealed_loader.go
dkharms: Yep, I recall that I wanted to do that as well but eventually forgot.

dkharms: Done.

PR #412, opened 2026-04-28
chore: don't evaluate skip masks in fetch after search

### LINKED ISSUE #411: Don't evaluate skip masks in fetch after search
If fetch is called as second phase after search, don't evaluate skip masks because ids have been already masked at search phase

forshev:
# Description

Add bool field to store's fetch request to not evaluate skip masks in fetch after search since these ids are already masked

Fixes #411

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```


api/storeapi/store_api.proto
```diff
@@ -254,6 +254,9 @@ message FetchRequest {
   bool explain = 3;
   repeated IdWithHint ids_with_hints = 4;
   FieldsFilter fields_filter = 5;
+  // if true, skip masks will be evaluated
+  // set to false in fetch after search since ids are already skipped
+  bool evalSkipMasks = 6;
```
eguguchkin: Although this is an internal API, let's still maintain backward compatibility and adjust it:

 - the new flag is optional
 - if not specified, the behavior remains unchanged
 - therefore, it should be named **no**SkipMask — i.e., **disable** the default behavior (filtering by skip mask)

eguguchkin (approved):

api/storeapi/store_api.proto
forshev: fixed

moflotas (approved):

PR #413, opened 2026-04-30
feat: implement merge source for compaction

dkharms:
# Description

Yet another pull request of series #336.
This one introduces all buildings blocks required for compaction of fractions.

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/compaction/merge.go
+++ b/compaction/merge.go
@@ -11,13 +11,13 @@ import (
 )
 
 func Merge(filename string, params common.SealParams, srcs ...Source) (*sealed.PreloadedData, error) {
-	writer := indexwriter.New(params)
+	w := indexwriter.New(params)
 	src := NewMergeSource(filename, srcs)
 
 	if err := createAndWrite(
 		filename+consts.OffsetsTmpFileSuffix,
 		filename+consts.OffsetsFileSuffix,
-		func(f *os.File) error { return writer.WriteOffsetsFile(f, src) },
+		func(f *os.File) error { return w.WriteOffsetsFile(f, src) },
 	); err != nil {
 		return nil, err
 	}
@@ -25,7 +25,7 @@ func Merge(filename string, params common.SealParams, srcs ...Source) (*sealed.P
 	if err := createAndWrite(
 		filename+consts.IDTmpFileSuffix,
 		filename+consts.IDFileSuffix,
-		func(f *os.File) error { return writer.WriteIDFile(f, src) },
+		func(f *os.File) error { return w.WriteIDFile(f, src) },
 	); err != nil {
 		return nil, err
 	}
@@ -35,7 +35,7 @@ func Merge(filename string, params common.SealParams, srcs ...Source) (*sealed.P
 		filename+consts.TokenFileSuffix,
 		filename+consts.LIDTmpFileSuffix,
 		filename+consts.LIDFileSuffix,
-		func(tf, lf *os.File) error { return writer.WriteTokenTriplet(tf, lf, src) },
+		func(tf, lf *os.File) error { return w.WriteTokenTriplet(tf, lf, src) },
 	); err != nil {
 		return nil, err
 	}
@@ -43,7 +43,7 @@ func Merge(filename string, params common.SealParams, srcs ...Source) (*sealed.P
 	if err := createAndWrite(
 		filename+consts.InfoTmpFileSuffix,
 		filename+consts.InfoFileSuffix,
-		func(f *os.File) error { return writer.WriteInfoFile(f, src) },
+		func(f *os.File) error { return w.WriteInfoFile(f, src) },
 	); err != nil {
 		return nil, err
 	}
@@ -69,13 +69,13 @@ func Merge(filename string, params common.SealParams, srcs ...Source) (*sealed.P
 		info.IndexOnDisk += uint64(st.Size())
 	}
 
-	lidsTable := writer.LIDsTable()
+	lidsTable := w.LIDsTable()
 	preloaded := &sealed.PreloadedData{
 		Info:       info,
-		TokenTable: writer.TokenTable(),
+		TokenTable: w.TokenTable(),
 		BlocksData: sealed.BlocksData{
 			LIDsTable:     &lidsTable,
-			IDsTable:      writer.IDsTable(),
+			IDsTable:      w.IDsTable(),
 			BlocksOffsets: src.BlockOffsets(),
 		},
 	}
@@ -90,7 +90,7 @@ func mergeDocs(filename string, srcs ...Source) error {
 		func(f *os.File) error {
 			var docsSize uint64
 			for _, src := range srcs {
-				for loc, err := range src.DocBlock() {
+				for loc, err := range src.DocBlocks() {
 					if err != nil {
 						return err
 					}
--- a/compaction/merge.go
+++ b/compaction/merge.go
@@ -1,0 +1,162 @@
+package compaction
+
+import (
+	"errors"
+	"os"
+
+	"github.com/ozontech/seq-db/consts"
+	"github.com/ozontech/seq-db/frac/common"
+	"github.com/ozontech/seq-db/frac/sealed"
+	"github.com/ozontech/seq-db/indexwriter"
+)
+
+func Merge(filename string, params common.SealParams, srcs ...Source) (*sealed.PreloadedData, error) {
+	writer := indexwriter.New(params)
+	src := NewMergeSource(filename, srcs)
+
+	if err := createAndWrite(
+		filename+consts.OffsetsTmpFileSuffix,
+		filename+consts.OffsetsFileSuffix,
+		func(f *os.File) error { return writer.WriteOffsetsFile(f, src) },
+	); err != nil {
+		return nil, err
+	}
+
+	if err := createAndWrite(
+		filename+consts.IDTmpFileSuffix,
+		filename+consts.IDFileSuffix,
+		func(f *os.File) error { return writer.WriteIDFile(f, src) },
+	); err != nil {
+		return nil, err
+	}
+
+	if err := createAndWriteBoth(
+		filename+consts.TokenTmpFileSuffix,
+		filename+consts.TokenFileSuffix,
+		filename+consts.LIDTmpFileSuffix,
+		filename+consts.LIDFileSuffix,
+		func(tf, lf *os.File) error { return writer.WriteTokenTriplet(tf, lf, src) },
+	); err != nil {
+		return nil, err
+	}
+
+	if err := createAndWrite(
+		filename+consts.InfoTmpFileSuffix,
+		filename+consts.InfoFileSuffix,
+		func(f *os.File) error { return writer.WriteInfoFile(f, src) },
+	); err != nil {
+		return nil, err
+	}
+
+	if err := mergeDocs(filename, srcs...); err != nil {
+		return nil, err
+	}
+
+	info := src.Info()
+	info.IndexOnDisk = 0
+
+	for _, suffix := range []string{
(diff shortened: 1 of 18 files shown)


compaction/merge_source.go
```diff
@@ -0,0 +1,445 @@
...
+	"github.com/ozontech/seq-db/frac/common"
+	"github.com/ozontech/seq-db/indexwriter"
+	"github.com/ozontech/seq-db/seq"
+	"github.com/ozontech/seq-db/util"
+)
+
+type (
+	Document         = util.Pair[seq.ID, []byte]
+	DocBlockLocation = util.Pair[[]byte, uint64]
+	TokenPosting     = util.Pair[[]byte, []uint32]
```
cheb0: I have a strange feeling that we start to use two terms for same thing (`lid` and `posting`). I'd stick to a single term (which is a `lid` since we don't have much choise). At the same time docs/comments might mention that lids are postings. 

Besides, a single posting == a single lid. So, `util.Pair[[]byte, []uint32]` is techically a posting list. But I'd suggest to stick to lids and rename to `TokenLids`

compaction/merge_source.go
```diff
@@ -0,0 +1,445 @@
...
+				token = c.posting.First
+			}
+		}
+
+		return token, set
+	}
+
+	// NB: This buffer will be reused across
+	// all calls within current field.
+	var lidRenamed []uint32
```
cheb0, line 377: nit: `lidsRenamed`?

frac/sealed_source.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (s *SealedSource) Info() *common.Info {
+	return s.f.info
+}
+
+func (s *SealedSource) BlockOffsets() []uint64 {
+	return s.f.blocksData.BlocksOffsets
+}
+
+func (s *SealedSource) ID() iter.Seq2[indexwriter.DocLocation, error] {
```
cheb0: nit: not that I care, but you use single noun func name which returns an iter. Is it some Go best practise?

cheb0 (approved):

frac/sealed_source.go
```diff
@@ -0,0 +1,159 @@
...
+					return
+				}
+			}
+		}
+	}
+}
+
+func (s *SealedSource) DocBlock() iter.Seq2[DocBlockLocation, error] {
+	return func(yield func(DocBlockLocation, error) bool) {
+		// We do not want to cache payload of DocBlock because
```
cheb0, line 140: Yes, but we would sill evict a lot from page cache. And page cache is used a lot by our search workers, it's just not that you can see it directly. So, for compaction (and write path) [direct I/O would be preferable](https://lightfoot.dev/direct-i-o-for-cassandra-compaction-cutting-p99-read-latency-by-5x). The only downside - compaction could also benefit from readahead.

There is also `sendfile` system call which will allow to bypass even userspace (zero copy).  It seems `io.Copy` directly uses `sendfile`. Might be a good choise for copying docblocks. However, it will not work if we unpack doc blocks.

dkharms: > Yes, but we would sill evict a lot from page cache

Unfortunately, we do not have low-level access methods otherwise I would use `fadvise(FADV_DONTNEED)`, of course.

But, I am not sure this will affect the page cache anyway. Linux is smart about page cache and eviction so such workloads with one-pass pattern will not affect active-list in page cache [1]

> direct I/O would be preferable

Yep, however it requires some guarantees that we cannot provide (alignment guarantees, for instance). And this requirement is stated in the article as well -- "It requires both disk operations and off-heap memory buffers to be aligned to the filesystem block size"

> There is also `sendfile` system call which will allow to bypass even userspace

Yep, I am aware of that. However, using `sendfile` for transferring data between files does not show noticeable performance improvements [2]:
```
sendfile             159ms  168ms  175ms
pipe+splice          161ms  162ms  163ms  (+1.3%)
read+write 16bs      164ms  165ms  178ms  (+3.1%)
```

and, again, we do not have necessary API for doing such low-level IO. On the other hand, there is one slightly different approach explored in this article with preallocating filesystem metadata with `fallocate` and it shows real performance gains.

And, AFAIK, `sendfile` was designed to efficient transferring between sockets, not files. So there's a better alternative `copy_file_range` [3]



---

[1] https://biriukov.dev/docs/page-cache/4-page-cache-eviction-and-page-reclaim/
[2] https://blog.plenz.com/2014-04/so-you-want-to-write-to-a-file-real-fast.html
[3] https://man7.org/linux/man-pages/man2/copy_file_range.2.html

compaction/merge_source.go
dkharms: Yeah, what I really meant here is `TokenToPostingList`, but for brevity I've dropped with part with `To` because this is `Pair`. Will rename it.

frac/sealed_source.go
dkharms: I do not know, honestly.

eguguchkin: Well, by the way, I agree with Andrey – "IDs" (plural) is more idiomatic for Go

frac/sealed_source.go
```diff
@@ -0,0 +1,159 @@
...
+func (s *SealedSource) postingsForField(field string) iter.Seq2[indexwriter.TokenPosting, error] {
+	lidsTable := s.f.blocksData.LIDsTable
+	tokenTable := s.tokenTableLoader.Load()
+
+	var lidsBuf []uint32
+	return func(yield func(indexwriter.TokenPosting, error) bool) {
+		for _, entry := range tokenTable[field].Entries {
+			block := s.tokenBlockLoader.Load(entry.BlockIndex)
+
+			for tid := entry.StartTID; tid < entry.StartTID+entry.ValCount; tid++ {
```
eguguchkin: nit: `for tid := entry.StartTID; tid <= entry.LastTID(); tid++ {`

frac/sealed_source.go
```diff
@@ -0,0 +1,159 @@
...
+
+				for bi := firstBlock; bi <= lastBlock; bi++ {
+					lidBlock, err := s.lidsLoader.GetLIDsBlock(bi)
+					if err != nil {
+						yield(indexwriter.TokenPosting{}, err)
+						return
+					}
+
+					chunkIdx := lidsTable.GetChunkIndex(bi, tid)
+					lidsBuf = append(lidsBuf, lidBlock.LIDs[lidBlock.Offsets[chunkIdx]:lidBlock.Offsets[chunkIdx+1]]...)
```
eguguchkin: nit: `lidsBuf = append(lidsBuf, lidBlock.getLIDs(chunkIdx)...)` and make getLIDs -> GetLIDs public?

frac/sealed_source.go
eguguchkin: > we do not have necessary API

I agree.  And let's leave such optimization for future PR, otherwise it turns out to be too complicated.

frac/sealed_source.go
```diff
@@ -0,0 +1,159 @@
...
+		for _, offset := range s.f.blocksData.BlocksOffsets {
+			// Read DocBlock payload (including its header) but do not decompress it.
+			// Caller of [SealedSource.DocBlock] will decide whether it requires decompressed data.
+			payload, _, err := r.ReadDocBlock(int64(offset))
+			if err != nil {
+				yield(DocBlockLocation{}, err)
+				return
+			}
+
+			loc := DocBlockLocation{First: payload, Second: offset}
```
eguguchkin, line 153: nit: Returning the offset seems redundant – we already obtain the payload higher up the stack, and we can locally advance the offset there as offset += len(payload).

In general, we need the writer's offset, but here we are passing the reader's offset; the fact that they coincide is purely a case‑specific detail of this implementation

compaction/merge_source.go
```diff
@@ -0,0 +1,445 @@
...
+	// and it is around log(k) vs 2*log(k).
+
+	type cursor struct {
+		next func() (string, iter.Seq2[TokenPosting, error], bool)
+		stop func()
+
+		field string
+		tokIt iter.Seq2[TokenPosting, error]
+
+		ok bool
```
eguguchkin, line 255: nit:  ok -> has

compaction/merge_source.go
```diff
@@ -0,0 +1,445 @@
...
+	// and it is around log(k) vs 2*log(k).
+
+	type cursor struct {
+		next func() (DocLocation, error, bool)
+		stop func()
+
+		loc    DocLocation
+		lidOld uint32
+
+		ok bool
```
eguguchkin, line 143: nit: ok -> has

compaction/merge_source.go
```diff
@@ -0,0 +1,445 @@
...
+
+		defer func() {
+			for _, c := range cursors {
+				c.stop()
+			}
+		}()
+
+		for i := range s.sources {
+			src := s.sources[i]
+			next, stop := iter.Pull2(src.ID())
```
eguguchkin: nit: 
```
next, stop := iter.Pull2(src.ID())
defer stop()
```

instead of 

```
defer func() {
	for _, c := range cursors {
		c.stop()
	}
}()
```

compaction/merge_source.go
```diff
@@ -0,0 +1,445 @@
...
+
+			// Rename lid from picked cursor to the new value.
+			s.lidMapping[idx][lidOld] = lid
+
+			var err error
+			c.loc, err, c.ok = c.next()
+			c.lidOld += 1
+
+			if err != nil {
+				cursors[idx] = c
```
eguguchkin: nit:

It looks redundant before exiting the function

compaction/merge_source.go
```diff
@@ -0,0 +1,445 @@
...
+				}
+			}
+
+			// All pull-iterators are exhausted.
+			// Close all iterators and return.
+			if idx == -1 {
+				break
+			}
+
+			c := cursors[idx]
```
eguguchkin, line 208: Theoretically, we could have collisions here – identical IDs in different factions. I would check that after each selection of the idx, it is not equal to the previous one; otherwise, it should be skipped, the cursor should be advanced, and in the LID map, that position should be marked as zero (so that it is also skipped in the lIDs list)

I don't want to overcomplicate things at this stage, but in the final implementation, in my opinion, this case needs to be handled.

compaction/merge.go
```diff
@@ -0,0 +1,162 @@
...
+		filename+consts.DocsFileSuffix,
+		func(f *os.File) error {
+			var docsSize uint64
+			for _, src := range srcs {
+				for loc, err := range src.DocBlock() {
+					if err != nil {
+						return err
+					}
+
+					payload, offset := loc.First, loc.Second
```
eguguchkin, line 98: nit: We don't need to calculate the offset from the source's offset here; we need to dynamically compute the offset during writing based on what we've written previously. In short, we don't need to pass the offset from the iterator

eguguchkin (approved):

frac/sealed_source.go
dkharms: Done.

frac/sealed_source.go
dkharms: Done.

compaction/merge_source.go
dkharms: Done.

frac/sealed_source.go
dkharms: Done.

compaction/merge_source.go
dkharms: It's not that simple because we have to correctly handle nested indexes -- in this case it is totally fine to have several identical ids in a row.

I'll leave a `TODO` here and will think about the implementation.

compaction/merge_source.go
dkharms: Done.

PR #417, opened 2026-05-04
feat: make fsync configurable via fraction config

### LINKED ISSUE #199: Make calling fsync configurable via fraction config
Currently, we can only disable `fsync` using configs for seq-db. Potentially, it would be great to be able to disable it via fraction config. This might be helpful in unit tests because running hundreds of them sometimes takes longer than it might take. Disable fsync can cut time in half.

dmedovich:
Fixes #199

Added `SkipFsync bool` to `frac.Config` and `common.SealParams`.
So fsync can be disabled at the fraction level, not just globally via `resources.skip_fsync`. 
Useful in unit tests — setting `frac.Config{SkipFsync: true}` cuts test time roughly in half when running hundreds of tests. Global `resources.skip_fsync` still works and now propagates into fraction config automatically on startup.

---

- [x] I have read and followed all requirements in CONTRIBUTING.md
- [ ] I used LLM/AI assistance to make this pull request

Model: -
Prompt: -


cheb0 (changes requested):
overall looks good to me. let's enable skip fsync feature in fraction_test.go

PR #421, opened 2026-05-06
perf(skip masks): skip masks as roaring bitmaps in fetch

### LINKED ISSUE #420: Use roaring bitmaps to eval skip masks in fetch
Use roaring bitmaps to eval skip masks in fetch

forshev:
# Description

Read skip masks as roaring bitmap for fetch

Fixes #420

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/skipmaskmanager/loader.go
+++ b/skipmaskmanager/loader.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 
+	"github.com/RoaringBitmap/roaring"
 	"go.uber.org/zap"
 
 	"github.com/ozontech/seq-db/cache"
@@ -20,12 +22,12 @@ type loader struct {
 	cashKey      uint32
 }
 
-func newLoader(filePath string, headersCache *cache.Cache[[]lidsBlockHeader]) (*loader, error) {
+func newLoader(filePath string, headersCache *cache.Cache[[]lidsBlockHeader]) *loader {
 	return &loader{
 		filePath:     filePath,
 		headersCache: headersCache,
 		cashKey:      hashFilePath(filePath),
-	}, nil
+	}
 }
 
 func (l *loader) getFile() (*os.File, error) {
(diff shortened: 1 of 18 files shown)


skipmaskmanager/skip_mask_manager.go
```diff
@@ -223,6 +220,44 @@ func (smm *SkipMaskManager) GetIDsIteratorByFrac(
 	return NewNMergedIterators(iterators), has, nil
 }
 
+// GetSkipMaskAsRoaringBitmap returns skip masks as roaring bitmap.
```
cheb0: nit: comment name doesn't match function name

cheb0 (approved):

skipmaskmanager/skip_mask_manager.go
```diff
@@ -223,6 +220,44 @@ func (smm *SkipMaskManager) GetIDsIteratorByFrac(
...
+	smm.fracsMu.RLock()
+	defer smm.fracsMu.RUnlock()
+
+	fracFiles, has := smm.fracs[fracName]
+	if !has {
+		return nil, nil
+	}
+
+	bitmap := roaring.New()
+	mu := &sync.Mutex{}
```
eguguchkin: nit: I can't say for sure whether using a mutex hurts performance and by how much, but I'd go with a non-blocking implementation here.
That is, I'd generate a separate bitmap in each goroutine and then merge them at the end.

skipmaskmanager/loader.go
```diff
@@ -99,46 +101,80 @@ func (l *loader) loadHeaders() ([]lidsBlockHeader, error) {
...
+			mu.Lock()
+			bitmap.Add(lid)
+			mu.Unlock()
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	if err := l.release(); err != nil {
```
eguguchkin: nit: We might exit the function on an error and never call release. In iterators, theoretically, we might also fail to reach release. For me, this is an open question: when is it better to close the file descriptor? Probably, it would be consistent to close the descriptor only when the faction is deleted — just like the index: we keep it open all the time. But that would require a more global refactoring. Alternatively, tie it to the lifetime of skipmaskmanager.loader, but then we would need to return its release() function together with the iterator.

```diff
diff --git a/frac/processor/search.go b/frac/processor/search.go
index 3c08eae0..53b2e906 100644
--- a/frac/processor/search.go
+++ b/frac/processor/search.go
@@ -95,7 +95,8 @@ func IndexSearch(
 	}
 
 	m = sw.Start("get_skip_lids")
-	skipLIDs, hasSkipLIDs, err := index.GetSkipLIDs(minLID, maxLID, params.Order.IsReverse())
+	skipLIDs, hasSkipLIDs, release, err := index.GetSkipLIDs(minLID, maxLID, params.Order.IsReverse())
+	defer release()
 	m.Stop()
 	if err != nil {
 		return nil, err
```

The comforting fact is that in Go, when an *os.File is garbage-collected, Close should be called (thanks to runtime.SetFinalizer), but this is not a reliable mechanism.

eguguchkin (approved):

skipmaskmanager/skip_mask_manager.go
forshev: fixed

skipmaskmanager/skip_mask_manager.go
forshev: done

skipmaskmanager/loader.go
forshev: made defered release in loadToBitmap

forshev: made GetSkipLIDs return `release` func

PR #425, opened 2026-05-14
perf: find `TokenTable` index once

dkharms:
# Description

`TableLoader.Load` previously re-scanned all token blocks on every call to find the start of the token table. This commit extracts that scan into `advanceToTable`, guarded by a `sync.Once`, so the linear scan over token blocks happens at most once per `TableLoader` instance.
                                                                                                                                

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/frac/sealed/token/table_loader.go
+++ b/frac/sealed/token/table_loader.go
@@ -3,6 +3,7 @@ package token
 import (
 	"encoding/binary"
 	"slices"
+	"sync"
 	"unsafe"
 
 	"go.uber.org/zap"
@@ -22,7 +23,9 @@ type TableLoader struct {
 	reader *storage.IndexReader
 	cache  *cache.Cache[Table]
 
-	i   uint32
+	once       sync.Once
+	tableIndex uint32
+
 	buf []byte
 }
 
@@ -46,6 +49,8 @@ func (l *TableLoader) Load() Table {
 			blocks []TableBlock
 			err    error
 		)
+
+		l.advanceToTable()
 
 		if l.isLegacy {
 			blocks, err = l.loadBlocksLegacy()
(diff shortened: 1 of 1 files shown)


forshev (approved):

frac/sealed/token/table_loader.go
```diff
@@ -161,6 +155,20 @@ func (l *TableLoader) loadBlocks() ([]TableBlock, error) {
 	return blocks, nil
 }
 
+func (l *TableLoader) advanceToTable() {
```
cheb0: Trying to understand how does it work.
1. we call `advanceToTable`. It advances to token table (incrementing `tokenIndex`) and does it only once.
2. we call `loadBlocks`, we iterate through blocks, still incrementing `tokenIndex`

Each step of those advances `tableIndex`. However, I see that we do call a cache `table, err := l.cache.GetWithError`. If cache is cleared, then we do not reset `tableIndex` because we do it only once and this `once` instance is linked to `TableLoader`.  

So, I'm just trying to make sure if this can happen? For me it's easier to check this in the following way:

```
func (l *TableLoader) Load() Table {
	l.cache.Evict(1)
	table, err := l.cache.GetWithError(CacheKeyTable, func() (Table, int, error) {
```

With master branch I get consistent results:
```
Search (cold) completed. Found 100. Total 0. Took: 5 ms.
Hot iteration 0 found:  100
Hot iteration 1 found:  100
```

With this branch search can't find anything:
```
Search (cold) completed. Found 0. Total 0. Took: 8 ms.
Hot iteration 0 found:  0
Hot iteration 1 found:  0
```

So, clearly something is changed in the loader behaviour, but I'm not sure if this can (realistically) happen.

dkharms: Yep, that a was bug. Thank you!

I've reworked this code so it does not rely on shared index.
So, `tableIndex` is calculated once and both `loadBlocks`/`loadBlocksLegacy` work with internal copies of that `tableIndex`.

cheb0 (approved):

PR #428, opened 2026-05-21
chore: entry count statistics for cache

cheb0:
# Description

allow to see how many blocks are in the cache right now on dashboards

---

### DIFF
--- a/cache/metrics.go
+++ b/cache/metrics.go
@@ -14,7 +14,6 @@ type Metrics struct {
 	SizeRead        prometheus.Counter
 	SizeOccupied    prometheus.Counter
 	SizeReleased    prometheus.Counter
-	EntriesOccupied prometheus.Counter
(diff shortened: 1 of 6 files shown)


cache/metrics.go
```diff
@@ -54,12 +56,14 @@ func (m *Metrics) reportMiss(size uint64, latencySec float64) {
 		m.MissTotal.Inc()
 		m.SizeOccupied.Add(float64(size))
 		m.MissLatency.Add(latencySec)
+		m.EntriesOccupied.Inc()
```
forshev: it seems like`EntriesOccupied` metric is identical to `MissTotal`

forshev (approved):

cache/metrics.go
cheb0: you are right. I deleted `EntriesOccupied` metric

eguguchkin (approved):

PR #429, opened 2026-05-22
feat: new api for compaction

dkharms:
# Description

In this PR I've implemented all necessary API for compaction. And also changed visibility of all methods inside `fracmanager` package since they should not be exported.

I expect that there will be slight tweaks and changes to this API (e.g. method signature change) but the whole idea will be the same.

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/fracmanager/fraction_registry.go
+++ b/fracmanager/fraction_registry.go
@@ -21,16 +21,17 @@ type fractionRegistry struct {
 
 	sealing    map[string]*syncAppender                 // fractions being sealed (0-5 typical)
 	sealed     PartitionedCollection[*refCountedSealed] // local sealed fractions (can be thousands)
+	compacting map[string]*refCountedSealed             // fractions participating in compaction
 	offloading PartitionedCollection[*refCountedSealed] // fractions being offloaded (0-5 typical)
 	remotes    PartitionedCollection[*refCountedRemote] // offloaded fractions (can be thousands)
 
 	stats registryStats // size statistics for monitoring
 
 	muAppender sync.RWMutex
-	appender   *syncAppender // currently active writable fraction
+	sappender  *syncAppender // currently active writable fraction
 
-	muAll sync.RWMutex
-	all   fractionsSnapshot // all fractions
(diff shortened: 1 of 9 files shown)


cheb0 (approved):

fracmanager/fraction_registry.go
```diff
@@ -380,14 +404,60 @@ func (r *fractionRegistry) PromoteToRemote(sealed *refCountedSealed, remote *fra
...
+		r.sealed.Del(name)
+		r.stats.sealed.Sub(s.Info())
+
+		r.compacting[name] = s
+		r.stats.compacting.Add(s.Info())
+
+		claimed = append(claimed, s)
+	}
+
 	r.rebuildSnapshot()
```
eguguchkin: No need to rebuild the snapshot here — we're just moving from one queue to another, so the overall set of fractions hasn't changed.

fracmanager/fraction_registry.go
```diff
@@ -355,23 +359,43 @@ loop:
...
 
-// PromoteToRemote moves fractions from offloading to remote queue when offloading completes.
+func (r *fractionRegistry) substituteWithSealed(produced *frac.Sealed, consumed ...*refCountedSealed) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	for _, f := range consumed {
+		info := f.Info()
+		r.stats.compacting.Sub(info)
+		delete(r.compacting, info.Name())
```
eguguchkin, line 391: Should we additionally check, for reliability, that such a fraction is indeed present in the compacting list?

eguguchkin (approved):

fracmanager/fraction_registry.go
dkharms: Yep, agree on this one.

fracmanager/fraction_registry.go
dkharms: Yeah, consistency check won't hurt -- I've added it.

Issue #92, opened 2025-08-25
Graceful degration when dealing with remote storages


dkharms:
For historical reasons, most of our functions for working with block fractions assumed they were working with a local file system.

For example, if you look at the code of the following functions and methods, you can see that any error triggers `panic()` or `logger.Fatal`, which leads either to catching the panic and returning an `HTTP 5XX` to the client, or to process termination:
- `Loader.Load`;
- `Loader.skipBlock`;

Working with S3 requires a different approach to error handling, since this component is less reliable than the local file system and is subject to more frequent errors.

Let's divide errors into two types: `soft-error` and `hard-error`.

In my opinion, `soft-error` type errors do NOT indicate that the data is corrupted or simply missing.
They indicate that some subsystem was unable to process the request correctly and we will need to retry it in the future.

For example, `soft-error` errors can include:
- absence of network connection (for any reason);
- various network timeouts;

`hard-error` type errors indicate that the data has been corrupted and further work with such data will lead to violation of invariants in the program.
For such types of errors, the only correct thing to do is to forcibly stop the application, raise an alert, write a log, and require manual intervention for manual investigation.

For example, `hard-error` errors can include:
- absence of a fraction file in S3, although it should be there;
- reading some corrupted file;

> The original definitions of "hard error" and "soft error" were for disks and go back to the early 1970s at least. A soft error was a recoverable error that could be handled by re-reading the disk sector. A hard error was a permanent disk error that could not be recovered.

Accordingly, it is proposed to think through this idea and implement it in the code.

PR #436, opened 2026-06-10
feat: store info in .remote file for remote shards to avoid extra S3

### LINKED ISSUE #435: Store .remote-info for remote shards to avoid extra S3 requests
Currently, for remote shards (uploaded to S3), only an empty file `<shard-name>.remote` exists locally. All `.info` data for such shards is retrieved either from `.frac_cache` (if cached) or directly from S3.

### Problems

1. **Cannot determine shard format without an S3 request**  
   We cannot tell whether a shard is a single index or split into multiple files without making a request to S3.

2. **Losing the only local source of `.info` when removing `.frac_cache`**  
   If we remove `.frac_cache`, every store initialization will require fetching `.info` from S3 for each remote shard. This is unacceptable in terms of speed and cost.

### Solution

For each remote shard, store a local file `<shard-name>.remote-info` – **in exactly the same format as `.info` for regular shards**.

**Legacy shard format indicator:**  
`info.BinaryDataVer < config.BinaryDataV3`

### Disk space estimate

- Average `.info` size: **450–500 bytes**
- One shard every 5 minutes (after compaction)  
  **Yearly volume:**  
  `500 bytes * (525,600 min / 5 min) ≈ 50 MB`

eguguchkin:
## Summary

  Remote fractions used to keep only an empty `.remote` marker locally, so on every
  startup seq-db had to talk to S3 to figure out what a fraction is: the loader ran an
  `Exists(.index)` check per remote fraction to detect legacy format, and the fraction
  Info was recovered from `.frac-cache` or downloaded from S3.

  This PR stores a copy of the fraction Info locally as a `<frac>.remote-info` file,
  created at offload time. Fractions offloaded with this version restore their Info
  with **zero S3 requests** at startup.

  ## Changes

  - **Offload (`frac.Sealed.Offload`)**
    - New format: `frac.info` is hard-linked to `frac._remote`, all index files + docs
      + info are uploaded to S3 in parallel, then `._remote` is atomically renamed to
      `.remote-info` (durable rename + directory fsync).
    - Legacy format: unchanged semantics, but the `.remote` marker is now created via
      durable touch + rename.
    - A leftover `._remote` from a failed attempt is safely reused on retry — a single
      transient S3 error no longer blocks offloading of the fraction until restart.
  - **Startup (`fracmanager`)**
    - Removed the per-remote-fraction `Exists(.index)` S3 check from `discover()`;
      legacy vs split format is now derived from `Info.BinaryDataVer`
      (`frac.IsSingleIndex()`), the `isLegacy` flag is gone from all constructors.
    - `._remote` files (interrupted offload) are detected in the manifest and cleaned
      up; fraction stays sealed and is offloaded again.
    - If fraction Info cannot be loaded from any source, the process now fails fast
      with a clear `logger.Fatal` instead of a nil-pointer panic later.
  - **Info loading (`frac.Remote.loadInfo`)** — 3-level fallback:
    1. local `.remote-info` (new, no S3)
    2. `.info` on S3 (older offloaded fractions)
    3. legacy `.index` on S3 (oldest fractions)
  - **Misc**
    - `common.Info.InfoOnDisk` (not persisted) so `IndexOnDisk` includes the info file
      size; `Remote.openDocs`/`openRemoteFile` reworked with explicit `mustExist`
      contract; durable fs helpers (`DurableHardLink`, `DurableRenameFile`,
      `DurableTouchFile`); `Offload` returns only `error`.

  ## Compatibility

  - No on-disk format change for S3 objects; `.info` is still uploaded as before.
  - Fractions offloaded by previous versions (empty `.remote` marker) keep working
    via the S3 fallbacks in `loadInfo`.

  ## Testing

  - New unit tests in `fracmanager/loader_test.go` cover discovery with `.remote-info`
    and all fallback scenarios (empty `.remote` + cache legacy/new, legacy fast path).

  ## Known TODOs (follow-ups)

  - Clean up orphaned S3 objects after a failed offload ("S3 zombies").
  - Startup still fails hard if S3 is unavailable for fractions without
    `.remote-info` and cache (see #92).

Fixes #435 

---

### DIFF
--- a/consts/consts.go
+++ b/consts/consts.go
@@ -85,7 +85,8 @@ const (
 	// We can remove it in the future releases.
 	IndexDelFileSuffix = ".index.del"
 
-	RemoteFractionSuffix = ".remote"
+	RemoteFractionSuffix     = ".remote"
+	RemoteFractionInfoSuffix = ".remote-info"
 
 	FracCacheFileSuffix = ".frac-cache"
 	CompactionPlan      = ".compaction-plan"
--- a/consts/consts.go
+++ b/consts/consts.go
@@ -85,8 +85,9 @@ const (
 	// We can remove it in the future releases.
 	IndexDelFileSuffix = ".index.del"
 
-	RemoteFractionSuffix     = ".remote"
-	RemoteFractionInfoSuffix = ".remote-info"
+	RemoteFractionSuffix        = ".remote"
+	RemoteFractionInfoSuffix    = ".remote-info"
+	RemoteFractionInfoTmpSuffix = "._remote-info"
 
 	FracCacheFileSuffix = ".frac-cache"
 	CompactionPlan      = ".compaction-plan"
--- a/frac/sealed.go
+++ b/frac/sealed.go
@@ -6,13 +6,13 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"sync"
 
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/ozontech/seq-db/cache"
+	"github.com/ozontech/seq-db/config"
 	"github.com/ozontech/seq-db/consts"
 	"github.com/ozontech/seq-db/frac/common"
 	"github.com/ozontech/seq-db/frac/processor"
@@ -39,8 +39,6 @@ type Sealed struct {
 	docsCache  *cache.Cache[[]byte]
 	docsReader storage.DocsReader
 
-	// IsLegacy is true for fractions that use the old single .index file format.
-	IsLegacy             bool
 	legacyFile           *os.File
 	legacyReaderProvider *storage.ReaderProvider
 
@@ -84,9 +82,8 @@ func NewSealed(
 	indexCache *IndexCache,
 	docsCache *cache.Cache[[]byte],
 	info *common.Info,
-	config *Config,
+	cfg *Config,
 	skipMaskProvider skipMaskProvider,
-	isLegacy bool,
 ) *Sealed {
 	f := &Sealed{
 		initMu: &sync.RWMutex{},
@@ -95,10 +92,9 @@ func NewSealed(
 		docsCache:   docsCache,
 		indexCache:  indexCache,
 
-		IsLegacy:     isLegacy,
 		info:         info,
 		BaseFileName: baseFile,
-		Config:       config,
+		Config:       cfg,
 
 		PartialSuicideMode: Off,
 
@@ -122,7 +118,7 @@ func NewSealedPreloaded(
 	rl *storage.ReadLimiter,
 	indexCache *IndexCache,
 	docsCache *cache.Cache[[]byte],
-	config *Config,
+	cfg *Config,
 	skipMaskProvider skipMaskProvider,
 ) *Sealed {
 	f := &Sealed{
@@ -137,7 +133,7 @@ func NewSealedPreloaded(
 
 		info:         preloaded.Info,
 		BaseFileName: baseFile,
-		Config:       config,
+		Config:       cfg,
 
 		skipMaskProvider: skipMaskProvider,
 	}
@@ -160,7 +156,11 @@ func NewSealedPreloaded(
 	return f
 }
 
-func (f *Sealed) openInfoLegacy() {
+func (f *Sealed) IsSingleIndex() bool {
+	return f.info.BinaryDataVer < config.BinaryDataV3
+}
+
+func (f *Sealed) openIndexLegacy() {
 	if f.legacyFile != nil {
 		return
 	}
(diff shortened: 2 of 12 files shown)


frac/sealed.go
dkharms: By the way, we can get rid of this awkward return signature `(bool, error)` and just return `error` because `fractionProvider` will always try to offload sealed fractions.

If you remember, I've added this boolean to distinguish between offloading attempts of active and sealed fractions.

frac/sealed.go
```diff
@@ -344,15 +354,18 @@ func (f *Sealed) Offload(ctx context.Context, u storage.Uploader) (bool, error)
...
 
-	remoteFracName := f.BaseFileName + consts.RemoteFractionSuffix
-	file, err := os.Create(remoteFracName)
+	info, err := io.ReadAll(f.infoFile)
+	if err != nil {
+		return true, err
+	}
+
+	remoteInfoName := f.BaseFileName + consts.RemoteFractionInfoSuffix
+	err = util.WriteFileAtomic(remoteInfoName, info, 0o666, ".tmp")
```
dkharms: Seems like if seq-db crashes (for some reason) in the middle of `util.WriteFileAtomic` and we end up with `seq-db-*.remote-info.tmp` -- it won't be able to boot.

I guess it's better to just perform rename from `.info` to `.remote-info`, right?

forshev (dismissed):

frac/sealed.go
eguguchkin: done

dkharms (approved):

forshev (approved):

fracmanager/frac_manifest.go
```diff
@@ -46,7 +50,15 @@ type fracManifest struct {
 
 // hasAllIndexFiles reports whether all 5 split index files are present.
 func (m *fracManifest) hasAllIndexFiles() bool {
-	return m.hasInfo && m.hasToken && m.hasOffsets && m.hasID && m.hasLID
+	return m.hasInfo && m.hasToken && m.hasOffsets && m.hasID && m.hasLID || m.hasIndex
```
forshev: nit: maybe wrap all the new files with parens like this `(m.hasInfo && m.hasToken && m.hasOffsets && m.hasID && m.hasLID) || m.hasIndex`? i find this more readable

frac/sealed.go
```diff
@@ -323,51 +315,87 @@ func (f *Sealed) init(full bool) {
...
-	if f.IsLegacy {
-		g.Go(func() error { return u.Upload(gctx, f.legacyFile) })
-	} else {
-		g.Go(func() error { return u.Upload(gctx, f.infoFile) })
-		g.Go(func() error { return u.Upload(gctx, f.tokenFile) })
-		g.Go(func() error { return u.Upload(gctx, f.offsetsFile) })
-		g.Go(func() error { return u.Upload(gctx, f.idFile) })
-		g.Go(func() error { return u.Upload(gctx, f.lidFile) })
+	g.Go(func() error {
+		infoFile, err := os.Open(f.BaseFileName + consts.InfoFileSuffix)
```
forshev, line 326: can you please explain why we need to open file before upload it?

forshev (approved):

frac/sealed.go
dkharms: Well, because previously we kept file descriptor open and the file descriptor itself was a part of `frac.Sealed`.

But not anymore. I am not sure I like it though -- like we deal with `.info` file only on startup and then we basically can forget about it.

frac/sealed.go
dkharms, line 179: You said that you wanted to remove assignment via callbacks.

consts/consts.go
```diff
@@ -85,7 +85,9 @@ const (
 	// We can remove it in the future releases.
 	IndexDelFileSuffix = ".index.del"
 
-	RemoteFractionSuffix = ".remote"
+	RemoteFractionSuffix     = ".remote"
+	RemoteFractionTmpSuffix  = "._remote"
+	RemoteFractionInfoSuffix = ".remote-info"
```
dkharms: I suggest to use only two extensions:
- `._remote` -- fraction is partially offloaded;
- `.remote` -- fraction is fully offloaded;

The action sequence will be the same:
- Create hardlink `._remote` to `.info`;
- Offload data and `rename` file `._remote` to `.remote`;
- All other fraction files (except `.remote` will be dropped);

The only issue is with legacy `.remote` files which do not contain info block but it is pretty easy to resolve:
```
(1) size = stat(.remote)
(2) if size != 0 => return load(.remote); 
(3) return fetch(.index) or fetch(.info);
```

In my opinion `.remote-info` is not necessary. And we already have to deal with `.remote` suffix anyway.

fracmanager/frac_manifest.go
eguguchkin: done

frac/sealed.go
eguguchkin: done

consts/consts.go
eguguchkin: done

PR #438, opened 2026-06-11
feat: add token frequencies

cheb0:
# Description

Adds token frequency which can be used to optimize query performance

---


forshev (approved):

config/config.go
```diff
@@ -76,6 +76,12 @@ type Config struct {
 			// BlockSize sets max lids (postings) saved per LIDs block.
 			BlockSize int `config:"block_size" default:"65536"`
 		} `config:"lids"`
+
+		Tokens struct {
+			// FreqThreshold specifies minimum number of lids (postings) a token should have
+			// so that frequency for that token will be stored inside token blocks.
+			FreqThreshold int `config:"freq_threshold" default:"50"`
```
eguguchkin: Nit: In the future, I would suggest evolving this approach by using a percentile‑based threshold instead of a fixed one. The distribution of posting sizes may vary across different datasets. Rather than empirically tuning this threshold, I propose specifying the percentile for which the hint should be applied.

From a performance perspective, we could apply the hint to all tokens, but that would be expensive in terms of memory. Thus, this parameter would indicate how much we are willing to increase the dataset size in order to enable hinting for a subset of queries. This percentile‑based threshold can be computed incrementally in the active fraction and used during sealing.

Additionally, I suggest storing this threshold per fraction (its absolute value may differ). If for fraction A the threshold is X, this would mean that if a token lacks a hint, its posting size is ≤ X.

consts/consts.go
```diff
@@ -23,6 +23,7 @@ const (
 
 	DefaultBulkRequestsLimit   = 32
 	DefaultSearchRequestsLimit = 32
+	DefaultTokenFreqThreshold  = 50
```
eguguchkin: Nit: I haven't checked what the difference would be between, say, 50 and 5.

frac/sealed/token/block_loader_test.go
```diff
@@ -0,0 +1,91 @@
...
+		FreqIndexes: []uint16{1},
+		Freqs:       []uint32{64},
+	}
+
+	var packBuf []uint32
+	packed := src.Pack(nil, packBuf)
+
+	var dst1, dst2 Block
+	require.NoError(t, dst1.Unpack(packed, config.BinaryDataV5, &UnpackBuffer{}))
+	require.NoError(t, dst2.Unpack(packed, config.BinaryDataV5, &UnpackBuffer{}))
```
eguguchkin: It seems we don't reuse the buffer here.

eguguchkin (approved):

packer/delta_bitpacker.go
```diff
@@ -36,6 +43,19 @@ func CompressDeltaBitpackUint32(dst []byte, values, buf []uint32) []byte {
 	return dst
 }
 
+// CompressDeltaBitpackUint16 uses a temporary buffer to copy and cast values from uint16 to uint32 so it's a bit slower than CompressDeltaBitpackUint32.
+func CompressDeltaBitpackUint16(dst []byte, values []uint16, buf []uint32) []byte {
```
eguguchkin, line 47: nit: maybe we should store it as uint32 in unpacked form after all, so we don't have this function with extra allocations.

frac/sealed/token/block_loader.go
```diff
@@ -120,11 +223,34 @@ func (l *BlockLoader) Load(index uint32) *Block {
...
+	decompressedUint32 []uint32 // temporary buffer for bitpack
+	decompressedUint16 []uint16 // temporary buffer for bitpack
+	compressed         []uint32 // temporary buffer for bitpack
+}
+
+func (b *UnpackBuffer) Reset(fracVer config.BinaryDataVersion) {
+	if fracVer < config.BinaryDataV5 {
+		return
+	}
+	if b.decompressedUint32 == nil {
```
eguguchkin, line 252: It is also necessary to reset decompressedUint16

cheb0: Nope, we didn't need to reset, since those slices are not reassigned. However I made it that we actualy reassign slices, i.e. instead of 

```
var freqs []uint32
_, freqs, err = packer.DecompressDeltaBitpackUint32(data, buf.decompressedUint32, buf.compressed)
b.Freqs = append(b.Freqs, freqs...)
```

we now have

```
_, buf.decompressedUint32, err = packer.DecompressDeltaBitpackUint32(data, buf.decompressedUint32, buf.compressed)
b.Freqs = append(b.Freqs, buf.decompressedUint32...)
```

I think it's now more readable and it's easier to understand the purpose of these buffers. I've added reset for both buffers.

At the same time I removed `compressed` buffer reset. It's not needed. The problem is potential multiple resizes, however I have a task for removing this buffer altogether (we can just cast `[]byte` to `[]uint32` if host uses littel endian byte order. I planned to do this initially but decided to postpone. So, soon this buffer will be removed.

frac/sealed/token/block_loader_test.go
cheb0: fixed

consts/consts.go
eguguchkin: What I meant was — have you actually checked what the dataset size difference would be between threshold 5 and threshold 50 on our real data? :)

cheb0: There is a difference, but very tiny. Thus, both default values `0.001%` and `0.005%` look good.

prod, docs total: 686513
freq_threshold_percentage: 0.000% (all tokens all freqs) => 29,408,250 bytes
freq_threshold_percentage: 0.001% (all tokens >= 7 docs) => 26,659,895 bytes
freq_threshold_percentage: 0.005% (all tokens >= 35 docs) => 26,298,201 bytes
freq_threshold_percentage: 100% (only 3 tokens have freqs) => 25,937,625 bytes

LB, docs total: 955826
freq_threshold_percentage: 0.000% (all tokens have freqs) =>  45,977,641 bytes
freq_threshold_percentage: 0.001% (all tokens >= 10 docs) =>  42,096,958 bytes
freq_threshold_percentage: 0.005% (all tokens >= 48 docs) => 41,816,016 bytes
freq_threshold_percentage: 100% (only 5 tokens have freqs) => => 41,708,292 bytes

config/config.go
cheb0: Great idea. Config prop now is `freq_threshold_percentage` which is specified in percents.

packer/delta_bitpacker.go
cheb0: possible solution, but i'm trying hard to reduce the memory usage for token blocks, since quite a few token blocks in cache might have freqs in them

PR #443, opened 2026-06-16
feat: add random search downsampling

### LINKED ISSUE #440: search: add random downsampling to accelerate result processing
When a search query returns millions of candidate documents, further processing becomes unnecessarily slow. In many cases, a smaller random sample (e.g., 1 000 instead of 1 000 000) preserves the statistical properties of the full result set while drastically reducing compute cost.

This PR introduces a Downsample parameter in SearchParams. If >1, the search pipeline randomly keeps only ~1/Downsample of LIDs before fetching MIDs/RIDs.

eguguchkin:
Fixes #440

---


frac/processor/search.go
```diff
@@ -209,118 +209,144 @@ func iterateEvalTree(
...
+		i := 0
+		for _, lid := range in {
+			if cnt%n == 0 {
+				in[i] = lid
+				i++
+			}
+			cnt++
+		}
+		return in[:i]
+	}
```
forshev, line 350: nit: maybe iterate over `in` with step equal to `n` when we need to downsample? this way the greater the `n`, the less iterations will be needed

something like this:

```go
return func(in []node.LID) []node.LID {
	writeIndex := 0
	for i := 0; i < len(in); i += n { // `n` is downsample coefficient
		in[writeIndex] = in[i]
		writeIndex++
	}
	return in[:writeIndex]
}
```

forshev (approved):

frac/processor/search.go
eguguchkin: In theory, we might have 100 blocks of 1 element each. Downsampling by 2 would never trigger inside such a function. That's why we employ a global stateful counter here, as it preserves the state between blocks.

I think some optimizations could be devised here, but your particular version won't work correctly

eguguchkin: For example, we could modify n at the end of the block to account for this on the first step of the next block. But let's leave that as an optimization for future PRs

frac/fraction_test.go
```diff
@@ -1912,6 +1912,272 @@ func (s *FractionTestSuite) TestFractionInfo() {
...
+	searchAndAssertIDs("downsample=0 (default)")
+
+	// downsample=0 explicitly — should return all documents
+	searchAndAssertIDs("downsample=0 (explicit)", withDownsample(0))
+
+	// downsample=1 — should return all documents
+	searchAndAssertIDs("downsample=1", withDownsample(1))
+}
+
+func (s *FractionTestSuite) TestSearchDownsampleWithAggAndHist() {
```
cheb0, line 2050: test  failed, next 5 runs were successful

Image

cheb0 (approved):

frac/fraction_test.go
eguguchkin: This is strange — I ran it 50 times in a row and everything passed successfully. However, I've increased the allowable error tolerance

PR #446, opened 2026-06-24
feat: add roaring bitmaps in posing lists

cheb0:
# Description

---

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -86,8 +86,8 @@ type Config struct {
 			BlockSize int `config:"block_size" default:"65536"`
 			// BitmapThreshold specifies minimum number of LIDs in the lid list
 			// which are serialized as bitmap. LIDs lists with more elements use bitmap encoding,
-			// while smaller lists use delta encoding.
-			BitmapThreshold int `config:"bitmap_threshold" default:"65536"`
+			// while smaller lists use delta encoding. Default value is 0 (disabled).
+			BitmapThreshold int `config:"bitmap_threshold"`
 		} `config:"lids"`
 	} `config:"sealing"`
 
--- a/config/config.go
+++ b/config/config.go
@@ -84,6 +84,10 @@ type Config struct {
 		Lids struct {
 			// BlockSize sets max lids (postings) saved per LIDs block.
 			BlockSize int `config:"block_size" default:"65536"`
+			// BitmapThreshold specifies minimum number of LIDs in the lid list
+			// which are serialized as bitmap. LIDs lists with more elements use bitmap encoding,
+			// while smaller lists use delta encoding.
+			BitmapThreshold int `config:"bitmap_threshold" default:"65536"`
 		} `config:"lids"`
 	} `config:"sealing"`
 
--- a/frac/sealed/lids/block.go
+++ b/frac/sealed/lids/block.go
@@ -61,7 +61,7 @@ type UnpackedBlock struct {
 //	[offsets: delta-bitpack []uint32]     — slice boundaries in the delta-encoded LIDs array
 //	[lids: delta-bitpack []uint32]        — concatenated delta-encoded LID values
 //
-// Each list i in [0, listsCount) is either a roaring bitmap (when i appears in bitmapIndexes)
+// Each ith list in [0, listsCount) is either a roaring bitmap (when i appears in bitmapIndexes)
 // or a delta-encoded slice lids[offsets[k]:offsets[k+1]], where k is its delta-encoded slot index.
 // Lists with length >= LidsBitmapThreshold are stored as bitmaps; shorter lists use delta-encoding.
 type Block struct {
--- a/frac/sealed/lids/block.go
+++ b/frac/sealed/lids/block.go
@@ -1,51 +1,282 @@
 package lids
 
 import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
 	"math"
+	"slices"
 	"unsafe"
 
+	"github.com/RoaringBitmap/roaring/v2"
+
 	"github.com/ozontech/seq-db/config"
+	"github.com/ozontech/seq-db/consts"
+	"github.com/ozontech/seq-db/node"
 	"github.com/ozontech/seq-db/packer"
 )
 
+const (
+	defaultLidsBitmapThreshold = math.MaxInt // bitmaps disabled by default
+)
+
+type BlockPacker struct {
+	buf                 []uint32 // bitpack buffer (reusable across packing)
+	bmIndexes           []uint32 // bitmap indexes (reusable across packing)
+	bitpackLIDs         []uint32
+	bitpackOffsets      []uint32
+	bm                  *roaring.Bitmap // reusable across packing
+	LidsBitmapThreshold int
+}
+
+func NewBlockPacker() *BlockPacker {
+	return &BlockPacker{
+		bm:                  roaring.NewBitmap(),
+		buf:                 make([]uint32, 0, consts.DefaultLIDBlockCap),
+		bitpackLIDs:         make([]uint32, 0, consts.DefaultLIDBlockCap),
+		bitpackOffsets:      make([]uint32, 0, consts.DefaultLIDBlockCap),
+		bmIndexes:           make([]uint32, 0, consts.DefaultLIDBlockCap),
+		LidsBitmapThreshold: defaultLidsBitmapThreshold,
+	}
+}
+
+// UnpackedBlock contains accumulated LIDs ready to pack. It's only used on sealing/compaction (index writing) time.
+type UnpackedBlock struct {
+	LIDs      []uint32
+	Offsets   []uint32
+	IsLastLID bool
+}
+
+// Block contains LIDs in variable format. It's used during search/queries processing.
+// Field types is used to distinguish format for every LID list. If it's positive or zero, then it's a slot index in offsets
+// (delta-encoding is used). If it's negative, then it's a bitmap slot index (index starts from -1, so -1 stands for bitmaps[0]).
+// If types is nil, then the entire block is delta-encoded.
+//
+// On-disk format:
+//
+//	[listsCount: uint32]     — number of LID lists in the block
+//	[bitmapsCount: uint32]   — number of lists stored as roaring bitmaps
+//	[bitmaps: bitmapsCount × roaring bitmap, serial format]
+//	[bitmapIndexes: delta-bitpack []uint32] — sorted list indices encoded as bitmaps
+//	[offsets: delta-bitpack []uint32]     — slice boundaries in the delta-encoded LIDs array
+//	[lids: delta-bitpack []uint32]        — concatenated delta-encoded LID values
+//
+// Each list i in [0, listsCount) is either a roaring bitmap (when i appears in bitmapIndexes)
+// or a delta-encoded slice lids[offsets[k]:offsets[k+1]], where k is its delta-encoded slot index.
+// Lists with length >= LidsBitmapThreshold are stored as bitmaps; shorter lists use delta-encoding.
 type Block struct {
-	LIDs    []uint32
-	Offsets []uint32
+	types   []int32           // determines LID list type: delta-encoded (non-negative value) or bitmap (negative value). nil for delta-encoded blocks
+	lids    []uint32          // all LIDs which are delta-encoded as a flat array
+	offsets []uint32          // offsets for delta-encoded LIDs
+	bitmaps []*roaring.Bitmap // all LIDs lists which are stored as bitmaps
+	lastLID bool              // legacy field, will be removed soon
 }
 
-func (b *Block) getCount() int {
-	return len(b.Offsets) - 1
+func (b *Block) GetCount() int {
+	if b.types != nil {
+		return len(b.types)
+	}
+
+	return len(b.offsets) - 1
 }
 
-func (b *Block) GetLIDs(i int) []uint32 {
-	return b.LIDs[b.Offsets[i]:b.Offsets[i+1]]
+func (b *Block) IsLastLID() bool {
+	return b.lastLID
 }
 
-func (b *Block) Pack(dst []byte, buf []uint32) []byte {
-	dst = packer.CompressDeltaBitpackUint32(dst, b.Offsets, buf)
-	dst = packer.CompressDeltaBitpackUint32(dst, b.LIDs, buf)
-	return dst
+func (b *Block) GetLIDs(i int) node.LIDBatch {
+	if b.types == nil {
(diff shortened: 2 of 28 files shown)


cheb0:
@seqbenchbot up 336-compaction mixed

cheb0:
@seqbenchbot down ec638c8c

fracmanager/sealer_test.go
dkharms, line 1: It's gonna be interesting to see how bitmaps will affect sealing latency (and system in overall):
```bash
benchstat ~/main.txt ~/bitmaps.txt
goos: linux
goarch: amd64
pkg: github.com/ozontech/seq-db/fracmanager
cpu: 12th Gen Intel(R) Core(TM) i5-12600K
                 │ /home/dkharms/main.txt │     /home/dkharms/bitmaps.txt      │
                 │         sec/op         │   sec/op     vs base               │
Sealing_NoSort-4              459.1m ± 2%   468.2m ± 2%  +1.97% (p=0.005 n=10)

                 │ /home/dkharms/main.txt │      /home/dkharms/bitmaps.txt       │
                 │          B/op          │     B/op      vs base                │
Sealing_NoSort-4             23.05Mi ± 0%   26.06Mi ± 0%  +13.05% (p=0.000 n=10)

                 │ /home/dkharms/main.txt │      /home/dkharms/bitmaps.txt      │
                 │       allocs/op        │  allocs/op   vs base                │
Sealing_NoSort-4              5.322k ± 0%   8.309k ± 0%  +56.11% (p=0.000 n=10)
```

frac/sealed/lids/block.go
```diff
@@ -1,69 +1,306 @@
...
+		return node.NewSliceBatch(b.lids[b.offsets[i]:b.offsets[i+1]])
+	}
+	t := b.types[i]
+	if t >= 0 {
+		return node.NewSliceBatch(b.lids[b.offsets[t]:b.offsets[t+1]])
+	}
+	return node.NewBitmapBatch(b.bitmaps[-t-1])
+}
+
+func (b *Block) CopyLIDs(idx int, dst []uint32) []uint32 {
```
dkharms: Maybe it's just me, but in Golang there is built-in `copy` function which overwrites `dst` but here we in fact do `append`, so I suggest to:

```suggestion
func (b *Block) AppendLIDsTo(idx int, dst []uint32) []uint32 {
```

frac/sealed/lids/block.go
```diff
@@ -1,69 +1,306 @@
...
+	}
+	t := b.types[idx]
+	if t >= 0 {
+		dst = append(dst, b.lids[b.offsets[t]:b.offsets[t+1]]...)
+		return dst
+	}
+	return b.copyLIDsFromBitmap(t, dst)
+}
+
+func (b *Block) copyLIDsFromBitmap(ref int32, buf []uint32) []uint32 {
```
dkharms:
```suggestion
func (b *Block) appendLIDsFromBitmapTo(ref int32, buf []uint32) []uint32 {
```

node/batch.go
dkharms, line 1: As I am getting deeper in this pull request I just have several questions about roaring bitmaps in general:
- Are they concurrent-safe for read-only use (getting max/min, cardinality)?
- Are they concurrent-safe for (possibly) mutating operations (like `Or(bm1, bm2)`) or there are different APIs for performing such operations in-place or returning a new copy instead?

frac/sealed/lids/block.go
```diff
@@ -1,69 +1,306 @@
...
+//	[offsets: delta-bitpack []uint32]     — slice boundaries in the delta-encoded LIDs array
+//	[lids: delta-bitpack []uint32]        — concatenated delta-encoded LID values
+//
+// Each list i in [0, listsCount) is either a roaring bitmap (when i appears in bitmapIndexes)
+// or a delta-encoded slice lids[offsets[k]:offsets[k+1]], where k is its delta-encoded slot index.
+// Lists with length >= LidsBitmapThreshold are stored as bitmaps; shorter lists use delta-encoding.
 type Block struct {
-	LIDs    []uint32
-	Offsets []uint32
+	types   []int32           // determines LID list type: delta-encoded (non-negative value) or bitmap (negative value). nil for delta-encoded blocks
```
dkharms: Have you measured whether materializing types actually buys anything vs. a binary search over `bitmapIndexes`? Like it's pretty easy to grasp but the liability of performing checks like `b.types == nil` is error-prone (IMHO).

And this decision is not scalable since you encode the type of block into signedness of integer (so there could be at most two different types). So if we are going to introduce different format for storing lids we would still have to rewrite this logic.

It could be easily rewritten as something like:
```go
func (b *Block) chunkRef(i int) (ChunkType, int) {
      k, found := slices.BinarySearch(b.bitmapIndexes, uint32(i))
      if found {
              return ChunkTypeBitmap, k
      }
      return ChunkTypeDelta, i - k
}
```

config/config.go
```diff
@@ -84,6 +84,10 @@ type Config struct {
 		Lids struct {
 			// BlockSize sets max lids (postings) saved per LIDs block.
 			BlockSize int `config:"block_size" default:"65536"`
+			// BitmapThreshold specifies minimum number of LIDs in the lid list
+			// which are serialized as bitmap. LIDs lists with more elements use bitmap encoding,
+			// while smaller lists use delta encoding.
+			BitmapThreshold int `config:"bitmap_threshold" default:"65536"`
```
dkharms: Should we add some kind of validation that `BitmapThreshold` cannot be larger than `BlockSize`?

frac/sealed/lids/block.go
```diff
@@ -1,69 +1,306 @@
...
+	size := int(unsafe.Sizeof(*b)) + uint32Size*cap(b.types) + uint32Size*cap(b.lids) + uint32Size*cap(b.offsets)
+	for _, bm := range b.bitmaps {
+		if bm != nil {
+			size += int(bm.GetSizeInBytes())
+		}
+	}
+	return size
+}
+
+func (p *BlockPacker) Pack(b *UnpackedBlock, dst []byte) []byte {
```
dkharms, line 130: Just curious. Have you tried to measure performance metrics when we include whole posting list for token when it exceeds 65k? Of course it heavily depends on density of lids but still it is quite interesting.

However it might break something if we use block capacity as a divider somewhere (I do not remember)...

node/node.go
```diff
@@ -14,9 +14,9 @@ type Node interface {
 type BatchedNode interface {
 	fmt.Stringer
 	// NextBatch returns next batch. Returns nil when exhausted.
-	NextBatch() LIDBatch
+	NextBatch(need int) LIDBatch
 	// NextBatchGeq returns next batch (LIDs >= minLID). Returns nil when exhausted.
-	NextBatchGeq(nextLID LID) LIDBatch
+	NextBatchGeq(need int, nextLID LID) LIDBatch
```
dkharms, line 19: Seems like `need` argument is not used anywhere. And in #479 it was deleted again.

dkharms (approved):

fracmanager/sealer_test.go
cheb0: So far I see they slow down quite literally everything except places where you actually intersect/union them :)

node/node.go
cheb0: Yes, I deleted it laster. As part of two PRs `need` will not be added.

frac/sealed/lids/block.go
```diff
@@ -1,69 +1,306 @@
...
+// On-disk format:
+//
+//	[listsCount: uint32]     — number of LID lists in the block
+//	[bitmapsCount: uint32]   — number of lists stored as roaring bitmaps
+//	[bitmaps: bitmapsCount × roaring bitmap, serial format]
+//	[bitmapIndexes: delta-bitpack []uint32] — sorted list indices encoded as bitmaps
+//	[offsets: delta-bitpack []uint32]     — slice boundaries in the delta-encoded LIDs array
+//	[lids: delta-bitpack []uint32]        — concatenated delta-encoded LID values
+//
+// Each list i in [0, listsCount) is either a roaring bitmap (when i appears in bitmapIndexes)
```
eguguchkin: typo?

> Each list i**S** in [0, listsCount) is either a roaring bitmap (when i**T** appears in bitmapIndexes)

frac/sealed/lids/block.go
```diff
@@ -1,69 +1,306 @@
...
+	t := b.types[idx]
+	if t >= 0 {
+		dst = append(dst, b.lids[b.offsets[t]:b.offsets[t+1]]...)
+		return dst
+	}
+	return b.copyLIDsFromBitmap(t, dst)
+}
+
+func (b *Block) copyLIDsFromBitmap(ref int32, buf []uint32) []uint32 {
+	bitmap := b.bitmaps[-ref-1]
```
eguguchkin: nit: It would be clearer if we passed the direct bitmap index to this function and performed the -t-1 calculation on the caller side.

`return b.copyLIDsFromBitmap(-t-1, dst)`

frac/sealed/lids/iterator_asc.go
```diff
@@ -1,44 +1,61 @@
...
+func NewIteratorAsc(
+	table *Table,
+	loader *Loader,
+	startIndex uint32,
+	tid uint32,
+	counter Counter,
+	minLID, maxLID uint32,
+) *IteratorAsc {
+	it := &IteratorAsc{
+		Cursor: *NewLIDsCursor(table, loader, startIndex, tid, counter, minLID, maxLID),
```
eguguchkin: nit: Perhaps we could avoid the pointer dereferencing complexity here?

```diff
--- a/frac/sealed/lids/cursor.go
+++ b/frac/sealed/lids/cursor.go
@@ -29,8 +29,8 @@ func NewLIDsCursor(
        tid uint32,
        counter Counter,
        minLID, maxLID uint32,
-) *Cursor {
-       return &Cursor{
+) Cursor {
+       return Cursor{
                table:  table,
                loader: loader,
```

frac/sealed/lids/iterator_asc.go
```diff
@@ -1,44 +1,61 @@
...
+func (it *IteratorAsc) narrowLIDsRange(tryNextBlock bool) bool {
+	if it.batch.IsEmpty() {
+		return tryNextBlock
 	}
 
-	last := lids[len(lids)-1]
-	if it.minLID > last { // fast path: out-of-bounds 2
-		return nil, false // stop reading blocks
+	first := it.batch.Min()
+	if it.maxLID < first {
```
eguguchkin: nit: We seem to have lost the comments here. It wasn't very clear even with them, but without them, it's quite hard to follow.

frac/sealed/lids/iterator_batched_asc.go
```diff
@@ -0,0 +1,94 @@
...
+		Cursor: it.Cursor,
+	}
+}
+
+func (*BatchedIteratorAsc) String() string {
+	return "LIDS_ASC_BATCHED"
+}
+
+// narrowLIDsRange cuts LIDs between minLID and maxLID. Returns updated tryNextBlock flag.
+func (it *BatchedIteratorAsc) narrowLIDsRange(tryNextBlock bool) bool {
```
eguguchkin, line 25: nit: It's the same implementation as in the regular iterator. We could just reuse a single shared one.

frac/sealed/lids/iterator_batched_asc.go
```diff
@@ -0,0 +1,94 @@
...
+	lastBlock := it.minLID > first
+	it.batch = it.batch.Narrow(it.minLID, it.maxLID)
+	if lastBlock {
+		tryNextBlock = false
+	}
+
+	return tryNextBlock
+}
+
+func (it *BatchedIteratorAsc) loadNextLIDsBlock() {
```
eguguchkin, line 51: nit: This implementation differs from the regular iterator by missing the following calls:

```
it.tryNextBlock = it.narrowLIDsRange(it.tryNextBlock)
it.counter.AddLIDsCount(it.batch.Len())
```
However, these are always called right alongside loadNextLIDsBlock.

So, maybe we should move them inside loadNextLIDsBlock? That way, both implementations would be unified with the regular iterator.

eguguchkin (approved):

node/batch.go
```diff
@@ -1,54 +1,326 @@
...
+	bm  *roaring.Bitmap
+	pos uint32
+}
+
+func newBitmapReverseIter(bm *roaring.Bitmap) *bitmapReverseIter {
+	return &bitmapReverseIter{bm: bm, pos: math.MaxUint32}
+}
+
+func (it *bitmapReverseIter) Next() (uint32, bool) {
+	prev := it.bm.PreviousValue(it.pos - 1)
```
eguguchkin, line 311: What if we use ReverseIterator inside Next()? Would it be faster?

frac/sealed/lids/block.go
eguguchkin: Doing a binary search on every Next() call is way too expensive, IMHO.

node/batch.go
cheb0: As far as I understand, they are thread-safe when not mutated. Currently, we clone before doing `Or` or `And`.

I decided to improve `fraction_concurrency_test` with a new test case for better guarantees.

frac/sealed/lids/block.go
cheb0: No not on every `Next`, but on every `Block.GetLIDs`. 

I decided to delete `types` array and go with binary search. We can bin search `bitmapIndexes`. If it's not found then we go to offsets.

config/config.go
cheb0: done

frac/sealed/lids/block.go
cheb0: I remember that I actually got better compression than current prod with 64k blocks (bitmaps were enabled for full-block posting lists). I got like `-1 Mb`. I still think we need block skips (skip pointers) tho, even for bitmaps, so not sure if we would be able to fully commit with this approach.

Besides, changing inverted index sort order is in plans. With a different sort order the results might be completely different.

frac/sealed/lids/block.go
cheb0: this comment doesn't exist now

frac/sealed/lids/iterator_asc.go
cheb0: restored comments

node/batch.go
cheb0: `ReverseIterator` doesn't have `advance` method

frac/sealed/lids/iterator_batched_asc.go
cheb0: fixed

PR #453, opened 2026-06-30
fix: add .compaction-plan to handle in-mid compaction crash

### LINKED ISSUE #336: Implement compaction of sealed fractions

dkharms:
# Description

Compaction's substitute-and-delete of stale fractions isn't atomic, so a crash mid-compaction could leave source fractions behind and produce duplicates on restart.

This adds a `.compaction-plan` file recording the participating fractions, which is written before merging and used on startup to clean up any leftover sources from an interrupted compaction.

Closes #336

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/compaction/merge.go
+++ b/compaction/merge.go
@@ -102,8 +102,8 @@ func createCompactionPlan(filename string, srcs ...Source) error {
 	// Later on, it will help us to make correct decision
 	// whether we should drop participants (delete them from disk).
 	return createAndWrite(
-		filename+consts.CompactionQueue,
-		filename+consts.CompactionQueue,
+		filename+consts.CompactionPlan,
+		filename+consts.CompactionPlan,
 		func(f *os.File) error {
 			var p plan
 
--- a/compaction/merge.go
+++ b/compaction/merge.go
@@ -1,6 +1,7 @@
 package compaction
 
 import (
+	"encoding/json"
 	"errors"
 	"os"
 
@@ -13,6 +14,10 @@ import (
 func Merge(filename string, params common.SealParams, srcs ...Source) (*sealed.PreloadedData, error) {
 	w := indexwriter.New(params)
 	src := NewMergeSource(filename, srcs)
+
+	if err := createCompactionPlan(filename, srcs...); err != nil {
+		return nil, err
+	}
 
 	if err := createAndWrite(
 		filename+consts.OffsetsTmpFileSuffix,
@@ -81,6 +86,37 @@ func Merge(filename string, params common.SealParams, srcs ...Source) (*sealed.P
 	}
 
 	return preloaded, nil
+}
+
+func createCompactionPlan(filename string, srcs ...Source) error {
+	type plan struct {
+		Participants []string `json:"participants"`
+	}
+
+	// Write which fractions are participating in compaction.
+	//
+	// Since we cannot remove all stale fractions and move
+	// the merged one into final state, we have to keep our
+	// intention on disk.
+	//
+	// Later on, it will help us to make correct decision
+	// whether we should drop participants (delete them from disk).
+	return createAndWrite(
+		filename+consts.CompactionQueue,
+		filename+consts.CompactionQueue,
+		func(f *os.File) error {
+			var p plan
+
+			for i := range srcs {
+				p.Participants = append(
+					p.Participants,
+					srcs[i].Info().Name(),
+				)
+			}
+
+			return json.NewEncoder(f).Encode(p)
+		},
+	)
 }
 
 func mergeDocs(filename string, srcs ...Source) error {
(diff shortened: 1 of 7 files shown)


dkharms:
@seqbenchbot up main bulk

dkharms:
@seqbenchbot down 8eedacf8

dkharms:
@seqbenchbot up main bulk

dkharms:
@seqbenchbot down 1c283919

dkharms:
@seqbenchbot up main bulk

dkharms:
@seqbenchbot down 36d0de9f

dkharms:
@seqbenchbot up main mixed

dkharms:
@seqbenchbot down 29d571ac

compaction/merge.go
```diff
@@ -83,6 +88,37 @@ func Merge(filename string, params common.SealParams, srcs ...Source) (*sealed.P
...
+	// Write which fractions are participating in compaction.
+	//
+	// Since we cannot remove all stale fractions and move
+	// the merged one into final state, we have to keep our
+	// intention on disk.
+	//
+	// Later on, it will help us to make correct decision
+	// whether we should drop participants (delete them from disk).
+	return createAndWrite(
+		filename+consts.CompactionQueue,
```
eguguchkin: here tmp = dst. Is there a risk of reading a partially written file?

fracmanager/frac_manifest.go
```diff
@@ -256,9 +277,63 @@ func filterValid(ids []string, manifests map[string]*fracManifest) ([]*fracManif
...
+
+			f := manifests[pid]
+			if f == nil {
+				// NOTE(dkharms): It is possible that compaction participants
+				// were dropped but the queue itself was not deleted.
+				continue
+			}
+
+			delete(filtered, pid)
+			removeAllFiles(f.basePath)
```
eguguchkin: nit: we need logs here for cases when we'll need to reconstruct the sequence from logs

fracmanager/frac_manifest.go
```diff
@@ -256,9 +277,63 @@ func filterValid(ids []string, manifests map[string]*fracManifest) ([]*fracManif
...
+			return nil, errors.New("inconsistent fraction file analysis")
+		}
+
+		skip := !m.hasCompactionQueue ||
+			m.Stage() != fracStageSealed
+
+		if skip {
+			continue
+		}
+
```
eguguchkin, line 317: nit: we need logs here on detecting an uncompleted compaction

fracmanager/frac_manifest.go
eguguchkin, line 277: nit: this can now be a valid state if compaction was interrupted, so

```go
			if !manifest.hasCompactionQueue {
				fractionLoadErrors.Inc()
			}
```

eguguchkin (approved):

compaction/merge.go
dkharms: Yep, forgot to handle that case. I do not want to introduce one more file extension which will require additional handling, so I will do something like:
```go
var p plan
if err := json.NewDecoder(f).Decode(&p); err != nil {
	// Well, we cannot decode compaction plan so let's drop whatever the result is.
	delete(filtered, id)
	removeAllFiles(m.basePath)
	continue
}
```

fracmanager/frac_manifest.go
dkharms: Yep, makes sense to distinguish this case.

fracmanager/frac_manifest.go
dkharms: It's redundant. Fix for this comment will be enough https://github.com/ozontech/seq-db/pull/453#discussion_r3558225895

fracmanager/frac_manifest.go
dkharms: Done.

compaction/merge.go
dkharms: Done.

fracmanager/frac_manifest.go
dkharms: Done.

cheb0 (approved):

PR #459, opened 2026-07-07
fix:  data loss in asyncsearch merge caused by concurrent retention deletion of fracs

eguguchkin:

### DIFF
--- a/asyncsearcher/async_searcher.go
+++ b/asyncsearcher/async_searcher.go
@@ -683,9 +683,10 @@ func (as *AsyncSearcher) FetchSearchResult(r FetchSearchResultRequest) (FetchSea
 
 	var qpr seq.QPR
 	var fracsDone, fracsInQueue int
+	histInterval := seq.MillisToMID(info.Request.Params.HistInterval)
 	if info.merged.Load() {
 		p := path.Join(as.config.DataDir, r.ID+asyncSearchExtMergedQPR)
-		qpr, _ = as.loadSearchResult([]string{p}, r.Limit, r.Order)
+		qpr, _ = as.loadSearchResult([]string{p}, r.Limit, r.Order, histInterval)
 		fracsDone = len(info.Fractions)
 		fracsInQueue = 0
 	} else {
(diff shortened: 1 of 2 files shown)


asyncsearcher/async_searcher.go
```diff
@@ -725,16 +726,21 @@ func (as *AsyncSearcher) FetchSearchResult(r FetchSearchResultRequest) (FetchSea
...
+func (as *AsyncSearcher) loadSearchResult(qprsPaths []string, limit int, order seq.DocsOrder, histInterval seq.MID) (seq.QPR, int) {
 	qpr := seq.QPR{}
 	size := 0
 	for _, qprPath := range qprsPaths {
 		compressedQPR, err := os.ReadFile(qprPath)
-		size += len(compressedQPR)
 		if err != nil {
+			if os.IsNotExist(err) {
+				// the corresponding fraction may have been removed concurrently by retention,
+				// so the qpr file may not exist
```
dkharms, line 737: Wait, but we are reading the `.qpr` file here which has different lifetime and does not depend on fraction's lifetime. Am I missing something?

Or you mean retention of `asyncsearcher` (like how long asynchronous request can survive)?

eguguchkin: Issue: when a user starts an async search with `from` close to the last fraction, the search may take hours. By the time it reaches that fraction, it might already be gone due to rotation. We currently ignore this as a valid case (https://github.com/ozontech/seq-db/blob/59fdb2b734cad2a10ce4801a68bdeb3676a3a754/asyncsearcher/async_searcher.go#L403). However, when we later merge all remaining fractions, a missing fraction's qpr fails the ENTIRE merge.

eguguchkin: Anyway, the async search queue will be reworked to use intervals soon, but for now we see this error a few times a week in logs. Users don't notice the missing data, but we should fix it sooner rather than later.

cheb0 (approved):

asyncsearcher/async_searcher.go
dkharms: > Anyway, the async search queue will be reworked to use intervals soon

I would not be so sure :)

dkharms (approved):

PR #471, opened 2026-07-23
chore(async search): search per N-minute intervals

### LINKED ISSUE #463: support compaction in async search
support compaction in async search

forshev:
# Description

search per N-minute intervals instead of per-fraction to support compaction

Fixes #463

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/asyncsearcher/async_searcher.go
+++ b/asyncsearcher/async_searcher.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/google/uuid"
 	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/ozontech/seq-db/bytespool"
 	"github.com/ozontech/seq-db/frac"
@@ -38,6 +39,8 @@ const (
 
 	minRetention = 5 * time.Minute
 	maxRetention = 30 * 24 * time.Hour // 30 days
+
+	defaultSearchInterval = 5 * time.Minute
 )
 
 type infoVersion uint8
@@ -45,7 +48,10 @@ type infoVersion uint8
 const (
 	infoVersion1 infoVersion = iota + 1 // MIDs stored in milliseconds
 	infoVersion2                        // MIDs stored in nanoseconds
+	infoVersion3                        // one file per N-minute interval instead of per fraction
 )
+
+type searchInterval [2]seq.MID
 
 type MappingProvider interface {
 	GetMapping() seq.Mapping
@@ -74,8 +80,7 @@ type AsyncSearcherConfig struct {
 }
 
 type fractionAcquirer interface {
-	AcquireFractions() (_ fracmanager.List, release func())
-	AcquireFraction(name string) (_ frac.Fraction, release func(), ok bool)
+	AcquireFractionsInRange(from, to seq.MID) (_ fracmanager.List, release func())
 }
 
 func MustStartAsync(config AsyncSearcherConfig, mp MappingProvider, fracProvider fractionAcquirer) *AsyncSearcher {
@@ -159,11 +164,7 @@ type asyncSearchInfo struct {
 	infoSize *atomic.Int64
 }
 
-func newAsyncSearchInfo(r AsyncSearchRequest, fracNames []string) asyncSearchInfo {
-	fracsToSearch := make([]fracSearchState, 0, len(fracNames))
-	for _, name := range fracNames {
-		fracsToSearch = append(fracsToSearch, fracSearchState{Name: name})
-	}
+func newAsyncSearchInfo(r AsyncSearchRequest) asyncSearchInfo {
 	ctx, cancel := context.WithCancel(context.Background())
 	return asyncSearchInfo{
 		Version:    infoVersion2,
@@ -173,7 +174,6 @@ func newAsyncSearchInfo(r AsyncSearchRequest, fracNames []string) asyncSearchInf
 		ctx:        ctx,
 		cancel:     cancel,
 		Request:    r,
-		Fractions:  fracsToSearch,
 		StartedAt:  time.Now(),
 		merged:     &atomic.Bool{},
 		qprsSize:   &atomic.Int64{},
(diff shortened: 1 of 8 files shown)


fracmanager/fraction_registry.go
```diff
@@ -100,6 +101,14 @@ func (r *fractionRegistry) acquireAllFractions() ([]frac.Fraction, func()) {
 	return r.snapshot.AcquireAll()
 }
 
+// acquireFractionsInRange returns a read-only subset of fractions within the range
+func (r *fractionRegistry) acquireFractionsInRange(from, to seq.MID) ([]frac.Fraction, func()) {
```
dkharms, line 105: By the way, we can reuse this API for regular fetches and searches. IIRC, now we acquire all fractions -- even those that do not intersect with user provided time range.

asyncsearcher/async_searcher.go
```diff
@@ -307,26 +307,53 @@ func (as *AsyncSearcher) createDataDir() {
...
+	splitInterval := seq.DurationToMID(defaultSearchInterval)
+	diff := to - from
+	return int((diff + splitInterval - 1) / splitInterval)
+}
+
+func intervalName(interval searchInterval) string {
+	return fmt.Sprintf("%d-%d", interval[0], interval[1])
+}
+
 func (as *AsyncSearcher) doSearch(id string, fracProvider fractionAcquirer) {
```
dkharms, line 344: One important thing to notice.

`compaction` and `asyncsearcher` can acquire the same fraction set. So, when both processes release fractions, `compaction` deletes merged fractions -- how result of asynchronous search will be returned in this case?

I understand that `QPR` is self-sufficient, but `QPR` contains `IDSources` which in their turn contain `Hint` (fraction name) which is used in `Fetch` stage.

fracmanager/fractions_snapshot.go
```diff
@@ -77,6 +78,33 @@ func (fs *fractionsSnapshot) AcquireAll() ([]frac.Fraction, func()) {
...
+	for i := range len(fs.fractions) {
+		f := fs.fractions[i]
+		c := fs.counters[i]
+
+		if f.Info().To < from {
+			continue
+		}
+		if f.Info().From > to {
+			break
+		}
```
dkharms: I think I've broken the invariant from comment in `fractionSnapshot` definition:
```go
fractions   []frac.Fraction // The actual fractions in chronological order
```

Now, we build snapshot in this way and it is totally possible that chronological order is broken:
```go
for s := range r.sealed.All() {
	all.AddSealed(s)
}

for _, c := range r.compacting {
	all.AddSealed(c)
}
```

So I suggest to remove this comment and filter fractions using the same approach as we have here:
```go
func (l List) FilterInRange(from, to seq.MID) List {
	res := make(List, 0, len(l))
	for _, f := range l {
		if f.IsIntersecting(from, to) {
			res = append(res, f)
		}
	}
	return res
}
```

eguguchkin: There has never been a chronological order here. Or rather, there used to be an order based on the time the faction was created, but not on 'From'.

asyncsearcher/async_searcher.go
```diff
@@ -307,26 +307,53 @@ func (as *AsyncSearcher) createDataDir() {
...
+	var intervals []searchInterval
+	current := from
+
+	for current < to {
+		end := min(current+splitInterval, to)
+		intervals = append(intervals, searchInterval{current, end - seq.MID(1)})
+		current = end
+	}
+
+	if len(intervals) > 0 {
```
eguguchkin, line 326: nit: I would just extend to++ before the loop and automatically get the correct intervals without needing to 'close last interval'

asyncsearcher/async_searcher.go
```diff
@@ -398,69 +448,102 @@ func compressQPR(qpr *seq.QPR, cb func(compressed []byte) error) error {
...
+		if panicData := util.RecoverToError(recover(), asyncSearchPanics); panicData != nil {
+			err = fmt.Errorf("async search panic during processFrac: %w", panicData)
+		}
+	}()
+
+	// use interval's from and to
+	info.Request.Params.From = interval[0]
+	info.Request.Params.To = interval[1]
+
+	qpr, err = f.Search(info.ctx, info.Request.Params)
```
eguguchkin: nit: just `return f.Search(info.ctx, info.Request.Params)` ?

fracmanager/fractions_snapshot.go
forshev: fixed

asyncsearcher/async_searcher.go
forshev: so i checked it, we won't fetch if we pass non-existing frac in hint, but it works with empty hint

asyncsearcher/async_searcher.go
eguguchkin, line 593: nit: is the same searchIdFromQPRPath(name) ?

eguguchkin, line 848: nit: This is no longer relevant now

fracmanager/fetcher.go
```diff
@@ -162,6 +167,12 @@ func groupIDsByFraction(idsOrig seq.IDSources, fracsIn List) (List, [][]seq.ID)
 
 			if id.Hint != fracName {
 				ids[i], i = id, i+1 // check this id for others fraction next time
+				if _, ok := fracsByName[id.Hint]; !ok {
```
eguguchkin: Let's just add a separate pass before the main loop

 ```go
	fracsByName := make(map[string]struct{}, len(fracsOut))
	for _, f := range fracsOut {
		fracsByName[f.Info().Name()] = struct{}{}
	}	
	for i, id := range ids {
		if id.Hint == "" {
			continue
		}
		if _, ok := fracsByName[id.Hint]; !ok {
			ids[i].Hint = ""
		}
	}
```

eguguchkin (approved):

fracmanager/fetcher.go
forshev: done

asyncsearcher/async_searcher.go
forshev: fixed fetch

dkharms (approved):

PR #473, opened 2026-07-24
feat(async search): add export async search docs method

### LINKED ISSUE #472: export async search result
We need to export documents found by async searches like regular search results

forshev:
# Description

Add a new method to export documents found by async search

Fixes #472

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```


api/seqproxyapi/v1/seq_proxy_api.proto
```diff
@@ -419,6 +427,12 @@ message ExportRequest {
   int64 offset = 3;      // Search offset.
 }
 
+message ExportAsyncSearchRequest {
+  string search_id = 1;
+  int64 size = 2;
+  int64 offset = 3;
```
dkharms, line 433: I guess you could an `order` here for documents order.

api/seqproxyapi/v1/seq_proxy_api.proto
```diff
@@ -419,6 +427,12 @@ message ExportRequest {
...
+}
```
dkharms, line 434: Do I understand correctly that `seq-ui` will advance offset here?

api/seqproxyapi/v1/seq_proxy_api.proto
```diff
@@ -112,6 +112,14 @@ service SeqProxyApi {
...
   }
+
+  // Stream documents for given async search ID.
+  rpc ExportAsyncSearch(ExportAsyncSearchRequest) returns (stream ExportResponse) {
+    option (google.api.http) = {
+      post: "/async-searches/export"
+      body: "*"
+    };
+  }
 }
```
dkharms, line 123: Sorry, but I do not understand the difference between `FetchAsyncSearchResult` and `ExportAsyncSearch`.

They both seem to return the documents and both call store's RPC `FetchAsyncSearchResult`.

Could you elaborate please?

dkharms (approved):

api/seqproxyapi/v1/seq_proxy_api.proto
```diff
@@ -419,6 +427,12 @@ message ExportRequest {
   int64 offset = 3;      // Search offset.
 }
 
+message ExportAsyncSearchRequest {
+  string search_id = 1;
+  int64 size = 2;
```
cheb0, line 432: nit: I can set `size` to a negative value and it panics inside. Would be nice to have some valiation. Same of `offset`

cheb0 (approved):

api/seqproxyapi/v1/seq_proxy_api.proto
forshev: yes

api/seqproxyapi/v1/seq_proxy_api.proto
forshev: ExportAsyncSearch is streaming

api/seqproxyapi/v1/seq_proxy_api.proto
forshev: fixed, thanks!

api/seqproxyapi/v1/seq_proxy_api.proto
forshev: we will take the original order of a search, it seems ok
also regular search export doesn't have order parameter so i didn't add it here to make the api as similar as posible

PR #479, opened 2026-07-29
perf: implement batch execution AND/OR/NAND

cheb0:
# Description

Batch execution allows to iterate over inverted index batch by batch instead of lid by lid. 

Batches are either of slice or roaring bitmap type. When batches are intersected/unioned they are converted to bitmaps.

### Enabling 

- query AST should suffice (AND/OR/AND NOT/range)
- `range` node can have at most 5 tids resolved
- iteration cost is evaluated and must be at least of configured number (freqs are required in fraction)

Iteration cost evaluation allows to enable batching when it's really worth it. In that case we know some good amount of LID blocks will be directly used as bitmaps and not be converted.

### Results

- **Searches** - I'd say most queries are unaffected except those which iterate over millions of postings.
 
<!DOCTYPE html>

<html>

<body>

Query | env | Total | cold, ms |   | hot, ms |   | cold (branch), ms |   | hot (branch), ms |   | cold diff | hot diff
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
service:large-service | prod | 220827 | 16.32 | ±3.45 | 0.54 | ±0.10 | 16.81 | ±6.27 | 0.7 | ±0.09 | 3% | 29.6%
service:large-service AND level:6 | prod | 111809 | 24.9 | ±3.25 | 6.75 | ±0.12 | 18.42 | ±1.76 | 2.56 | ±0.49 | -26% | -62.1%
service:small-service AND level:3 | prod | 349 | 35.75 | ±1.03 | 2.07 | ±0.13 | 38.12 | ±1.89 | 2.37 | ±0.12 | 6.6% | 14.5%
service:srv* AND level:[0 to 3] | prod | 214 | 51.54 | ±1.94 | 8.49 | ±0.15 | 51.67 | ±1.54 | 9.43 | ±0.19 | 0.3% | 11.1%
request_host:large-host AND response_status:200 | lb | 1058473 | 47.01 | ±2.74 | 21.16 | ±0.36 | 22.23 | ±1.69 | 5.09 | ±0.25 | -52.7% | -75.9%
request_host:large-host AND response_status:500 | lb | 2 | 14.91 | ±0.33 | 0.31 | ±0.03 | 16.62 | ±2.67 | 0.48 | ±0.15 | 11.5% | 54.8%
request_host:large-host AND request_method:POST AND geoip_country:RU | lb | 485486 | 66.71 | ±5.18 | 30.11 | ±0.17 | 25.66 | ±2.96 | 5.18 | ±0.70 | -61.5% | -82.8%
(response_status:504 or response_status:502) and (cluster_name:cl1 or cluster_name:cl2) | lb | 2996 | 49.42 | ±2.28 | 2.2 | ±0.06 | 44.76 | ±2.00 | 1.66 | ±0.03 | -9.4% | -24.5%
k8s_service_name:medium-service and request_host:medium-host | lb | 357675 | 21.1 | ±0.66 | 5.31 | ±0.13 | 17.55 | ±1.31 | 2.02 | ±0.10 | -16.8% | -62%
(request_host:small-host or request_host:small-host2) and hostname:some-server | lb | 25 | 19.72 | ±0.69 | 0.42 | ±0.03 | 18.66 | ±0.73 | 0.45 | ±0.04 | -5.4% | 7.1%
NOT response_status:200 | lb | 192352 | 32.85 | ±3.50 | 13.19 | ±0.20 | 45.88 | ±2.08 | 27.34 | ±0.72 | 39.7% | 107.3%
request_host:large-host AND NOT response_status:200 | lb | 6747 | 69.38 | ±2.03 | 17.32 | ±0.15 | 22.18 | ±2.43 | 4.22 | ±1.12 | -68% | -75.6%
request_host:small-host AND NOT (request_method:post or request_method:get) | lb | 0 | 43.58 | ±2.40 | 17.62 | ±0.26 | 17.32 | ±0.75 | 2.77 | ±0.25 | -60.3% | -84.3%

</body>

</html>

- **Histograms** - behave like searches but improvement is lower since we read MIDs
<!DOCTYPE html>

<html>
<body>

Query | env | Total | cold, ms |   | hot, ms |   | cold (branch), ms |   | hot (branch), ms |   | cold diff | hot diff
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
service:large-service AND level:[4 to 6] \| hist 60s | prod | 218423 | 95.5 | ±3.71 | 12.59 | ±0.41 | 87.72 | ±5.60 | 6.97 | ±1.43 | -8.1% | -44.6%
request_host:large-host AND response_status:200 \| hist 60s | lb | 1058473 | 136.62 | ±5.49 | 27.33 | ±1.49 | 112.97 | ±4.36 | 11.1 | ±0.98 | -17.3% | -59.4%
</body>
</html>

- **Aggs** - Mostly aggs do not use batching, but there is an overhead from allocating batches in LID cursors.
<!DOCTYPE html>

<html>

<body>

Query | env | Total | cold, ms |   | hot, ms |   | cold (branch), ms |   | hot (branch), ms |   | cold diff | hot diff
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
\|service:marketing-actions-api \| by k8s_pod | prod | 112107 | 256.22 | ±4.80 | 173.08 | ±3.03 | 294.39 | ±29.50 | 216.32 | ±24.66 | 14.9% | 25%
request_host:api.ozon.ru \| by remote_addr | lb | 1065220 | 2307.86 | ±107.43 | 2148.07 | ±127.33 | 2830.23 | ±473.64 | 2591.46 | ±388.58 | 22.6% | 20.6%

</body>

</html>

### Major problems
There are problems I found while working on batch execution. Can be partially addressed in future.
- `LIDBatch` - interface dispatch overhead is now present on inverted index which has partially affected hot queries performance. 
- roaring bitmap doesn't have `NextGEQ` (single function)
- Batches can grow very large, i.e. level:[5 to 7] (`nodeOrBatchedMulti`) can yield a large batch of size more than LID block. Truncating a batch can cost CPU and increase query execution time.
- Batches needs to be allocated
- Batch truncation takes CPU time. Current execution model is same as for LIDs which means we need truncation when do OR/AND NOT. It's possible to redo the model with full materialization. It will sacrifice early exit and block skipping but can be more performant with current inverted index sort order.

---

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -176,6 +176,15 @@ type Config struct {
 		} `config:"aggregation"`
 	} `config:"limits"`
 
+	QueryOptimization struct {
+		BatchExecution struct {
+			Enabled bool `config:"enabled"`
+			// CostThreshold is the minimum estimated non-batched execution cost required to enable batch-at-a-time query
+			// evaluation. Suggestion is to use value which is greater than 3 x LID block size.
+			CostThreshold int `config:"cost_threshold" default:"50000"`
+		} `config:"batch_execution"`
+	} `config:"query_optimization"`
+
 	CircuitBreaker struct {
 		Bulk struct {
 			// Checkout [CircuitBreaker] for more information.
--- a/config/config.go
+++ b/config/config.go
@@ -181,7 +181,7 @@ type Config struct {
 			Enabled bool `config:"enabled"`
 			// CostThreshold is the minimum estimated non-batched execution cost required to enable batch-at-a-time query
 			// evaluation. Suggestion is to use value which is greater than 3 x LID block size.
-			CostThreshold int `config:"cost_threshold" default:"50000"`
+			CostThreshold int `config:"cost_threshold" default:"150000"`
 		} `config:"batch_execution"`
 	} `config:"query_optimization"`
 
--- a/frac/processor/batch_eval_tree.go
+++ b/frac/processor/batch_eval_tree.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/ozontech/seq-db/config"
 	"github.com/ozontech/seq-db/metric/stopwatch"
 	"github.com/ozontech/seq-db/node"
 	"github.com/ozontech/seq-db/parser"
@@ -22,6 +23,7 @@ type leafTIDsCache map[parser.Token][]uint32
 // Returns errBatchingUnsupported when the non-batched path should be used.
 func tryBuildBatchEvalTree(
 	root *parser.ASTNode,
+	fracVer config.BinaryDataVersion,
 	ti tokenIndex,
 	queryOpts QueryOptimizationConfig,
 	minLID, maxLID uint32,
@@ -30,6 +32,10 @@ func tryBuildBatchEvalTree(
 	sw *stopwatch.Stopwatch,
 ) (node.BatchedNode, error) {
 	if !queryOpts.BatchExecution.Enabled {
+		return nil, errBatchingUnsupported
+	}
+	if fracVer < config.BinaryDataV6 {
+		// block batching for earlier versions (avoid delta-encoded posting lists converted to bitmaps)
 		return nil, errBatchingUnsupported
 	}
 
--- a/frac/processor/batch_eval_tree.go
+++ b/frac/processor/batch_eval_tree.go
@@ -60,17 +60,13 @@ func astSupportsBatching(root *parser.ASTNode) bool {
 		return false
 	}
 
-	switch root.Value.(type) {
+	switch v := root.Value.(type) {
 	case *parser.Literal:
-		literal := root.Value.(*parser.Literal)
-		// currently batching supports only simple terms like 'field:A'
-		return len(literal.Terms) == 1 && literal.Terms[0].Kind == parser.TermText
+		return len(v.Terms) == 1 && v.Terms[0].Kind == parser.TermText
 	case *parser.Range:
 		return true
 	case *parser.Logical:
-		logical := root.Value.(*parser.Logical)
-		// batching is not supported for NOT nodes yet
-		if logical.Operator == parser.LogicalNot {
+		if v.Operator == parser.LogicalNot {
 			return false
 		}
 		for i := range root.Children {
@@ -129,6 +125,8 @@ func leafIterationCost(ti tokenIndex, field string, token parser.Token, cache le
 	if err != nil {
 		return 0, err
 	}
+
+	// Currently we do not support batches for queries with deep trees. For example, queries like 'service:abc* AND level:*'
 	if len(tids) > maxBatchedTIDsPerLeaf {
 		return 0, errBatchingUnsupported
 	}
--- a/frac/processor/batch_eval_tree.go
+++ b/frac/processor/batch_eval_tree.go
@@ -1,0 +1,223 @@
+package processor
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/ozontech/seq-db/metric/stopwatch"
+	"github.com/ozontech/seq-db/node"
+	"github.com/ozontech/seq-db/parser"
+	"github.com/ozontech/seq-db/seq"
+)
+
+var errBatchingUnsupported = errors.New("batching unsupported")
+
+// maxBatchedTIDsPerLeaf limits number of TIDs for a single OrMulti node
+const maxBatchedTIDsPerLeaf = 5
+
+type leafTIDsCache map[parser.Token][]uint32
+
+// tryBuildBatchEvalTree tries to build a batched eval tree if possible.
+//
+// Returns errBatchingUnsupported when the non-batched path should be used.
+func tryBuildBatchEvalTree(
+	root *parser.ASTNode,
+	ti tokenIndex,
+	queryOpts QueryOptimizationConfig,
+	minLID, maxLID uint32,
+	stats *searchStats,
+	order seq.DocsOrder,
+	sw *stopwatch.Stopwatch,
+) (node.BatchedNode, error) {
+	if !queryOpts.BatchExecution.Enabled {
+		return nil, errBatchingUnsupported
+	}
+
+	if !astSupportsBatching(root) {
+		return nil, errBatchingUnsupported
+	}
+
+	cache := make(leafTIDsCache)
+	cost, err := calculateQueryIterationCost(root, ti, cache)
+	if err != nil {
+		return nil, err
+	}
+
+	threshold := queryOpts.BatchExecution.CostThreshold
+	if threshold <= 0 || cost <= uint64(threshold) {
+		return nil, errBatchingUnsupported
+	}
+
+	return buildBatchEvalTree(root, minLID, maxLID, stats, order.IsDesc(),
+		func(token parser.Token) (node.BatchedNode, error) {
+			return evalBatchLeaf(ti, token, cache, sw, stats, minLID, maxLID, order)
+		},
+	)
+}
+
+func astSupportsBatching(root *parser.ASTNode) bool {
+	if root == nil {
+		return false
+	}
+
+	switch root.Value.(type) {
+	case *parser.Literal:
+		literal := root.Value.(*parser.Literal)
+		// currently batching supports only simple terms like 'field:A'
+		return len(literal.Terms) == 1 && literal.Terms[0].Kind == parser.TermText
+	case *parser.Range:
+		return true
+	case *parser.Logical:
+		logical := root.Value.(*parser.Logical)
+		// batching is not supported for NOT nodes yet
+		if logical.Operator == parser.LogicalNot {
+			return false
+		}
+		for i := range root.Children {
+			if !astSupportsBatching(root.Children[i]) {
+				return false
+			}
+		}
+		return true
+	default:
+		return false
+	}
+}
+
+func calculateQueryIterationCost(root *parser.ASTNode, ti tokenIndex, cache leafTIDsCache) (uint64, error) {
+	if root == nil {
+		return 0, fmt.Errorf("empty AST")
+	}
+
+	switch token := root.Value.(type) {
+	case *parser.Literal:
+		return leafIterationCost(ti, token.Field, token, cache)
+	case *parser.Range:
+		return leafIterationCost(ti, token.Field, token, cache)
+	case *parser.Logical:
+		if len(root.Children) == 0 {
+			return 0, nil
+		}
+		childCosts := make([]uint64, len(root.Children))
+		for i, child := range root.Children {
+			c, err := calculateQueryIterationCost(child, ti, cache)
+			if err != nil {
+				return 0, err
+			}
+			childCosts[i] = c
+		}
+		if len(childCosts) != 2 {
+			return 0, fmt.Errorf("logical operator has unsupported count of children: %d", len(childCosts))
+		}
+		switch token.Operator {
+		case parser.LogicalAnd:
+			return min(childCosts[0], childCosts[1]), nil
+		case parser.LogicalNAnd:
+			return childCosts[0] + childCosts[1], nil
+		case parser.LogicalOr:
+			return childCosts[0] + childCosts[1], nil
+		default:
+			return 0, fmt.Errorf("unsupported logical operator for cost estimation: %v", token.Operator)
+		}
+	default:
+		return 0, fmt.Errorf("unsupported token type for cost estimation")
+	}
+}
+
+func leafIterationCost(ti tokenIndex, field string, token parser.Token, cache leafTIDsCache) (uint64, error) {
+	tids, err := ti.GetTIDsByTokenExpr(token)
+	if err != nil {
+		return 0, err
+	}
+	if len(tids) > maxBatchedTIDsPerLeaf {
+		return 0, errBatchingUnsupported
+	}
+	cache[token] = tids
+	if len(tids) == 0 {
+		return 0, nil
+	}
+
+	freqs := ti.GetFreqsByTIDs(tids, field)
+	var cost uint64
+	for _, freq := range freqs {
+		cost += uint64(freq)
+	}
+	return cost, nil
+}
+
+// buildBatchEvalTree builds a BatchedNode eval tree using already-validated leaf TIDs.
+func buildBatchEvalTree(
+	root *parser.ASTNode,
+	minLID, maxLID uint32,
+	stats *searchStats,
+	desc bool,
+	newBatchLeaf func(parser.Token) (node.BatchedNode, error),
+) (node.BatchedNode, error) {
+	if root == nil {
+		return nil, fmt.Errorf("empty AST")
+	}
+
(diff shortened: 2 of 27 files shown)


frac/sealed_index.go
```diff
@@ -275,6 +278,24 @@ func (ti *sealedTokenIndex) GetTIDsByTokenExpr(t parser.Token) ([]uint32, error)
...
 
+func (ti *sealedTokenIndex) GetFreqsByTIDs(tids []uint32, field string) []uint32 {
+	freqs := make([]uint32, len(tids))
+	if len(tids) == 0 {
+		return freqs
+	}
+
+	tokenTable := ti.tokenTableLoader.Load()
+	for i, tid := range tids {
+		if tid == 0 {
```
eguguchkin, line 289: nit: This shouldn't really happen. If it does, there's a bug somewhere. Better to panic here than to silently ignore it, right?

frac/sealed_index.go
```diff
@@ -275,6 +278,24 @@ func (ti *sealedTokenIndex) GetTIDsByTokenExpr(t parser.Token) ([]uint32, error)
 	return tids, nil
 }
 
+func (ti *sealedTokenIndex) GetFreqsByTIDs(tids []uint32, field string) []uint32 {
```
eguguchkin, line 281: nit: It seems like it would be more optimal to fetch both the token and its frequency in a single block load inside GetTIDsByTokenExpr() — but that's more complicated than fetching them separately.

frac/sealed_index.go
```diff
@@ -306,6 +327,37 @@ func (ti *sealedTokenIndex) GetLIDsFromTIDs(tids []uint32, stats lids.Counter, m
...
 }
 
+func (ti *sealedTokenIndex) GetBatchedLIDsFromTIDs(tids []uint32, stats lids.Counter, minLID, maxLID uint32, order seq.DocsOrder) []node.BatchedNode {
+	var (
+		getBlockIndex          func(tid uint32) uint32
+		getBatchedLIDsIterator func(uint32, uint32) node.BatchedNode
+	)
+
+	if order.IsReverse() {
+		getBlockIndex = func(tid uint32) uint32 { return ti.lidsTable.GetLastBlockIndexForTID(tid) }
```
eguguchkin: лишняя вложенная функция

```diff
--- a/frac/sealed_index.go
+++ b/frac/sealed_index.go
@@ -303,12 +303,12 @@ func (ti *sealedTokenIndex) GetLIDsFromTIDs(tids []uint32, stats lids.Counter, m
        )

        if order.IsReverse() {
-               getBlockIndex = func(tid uint32) uint32 { return ti.lidsTable.GetLastBlockIndexForTID(tid) }
+               getBlockIndex = ti.lidsTable.GetLastBlockIndexForTID
                getLIDsIterator = func(startIndex uint32, tid uint32) node.Node {
                        return lids.NewIteratorAsc(ti.lidsTable, ti.lidsLoader, startIndex, tid, stats, minLID, maxLID)
                }
        } else {
-               getBlockIndex = func(tid uint32) uint32 { return ti.lidsTable.GetFirstBlockIndexForTID(tid) }
+               getBlockIndex = ti.lidsTable.GetFirstBlockIndexForTID
                getLIDsIterator = func(startIndex uint32, tid uint32) node.Node {
                        return lids.NewIteratorDesc(ti.lidsTable, ti.lidsLoader, startIndex, tid, stats, minLID, maxLID)
                }
```

node/batch.go
```diff
@@ -111,17 +115,41 @@ func (b *sliceBatch) ReverseIter() Iter {
...
+	return it
+}
+
+type sliceManyIter struct {
+	lids []uint32
+	pos  int
+	desc bool
+}
+
+func (it *sliceManyIter) CopyLIDs(dst []LID, tmp []uint32) int {
```
eguguchkin: tmp не используется

node/batch.go
```diff
@@ -227,19 +255,43 @@ func (b *bitmapBatch) ReverseIter() Iter {
...
+
+type bitmapManyIterAsc struct {
+	it roaring.ManyIntIterable
+}
+
+func (it *bitmapManyIterAsc) CopyLIDs(dst []LID, tmp []uint32) int {
+	if len(dst) == 0 || len(tmp) == 0 {
+		return 0
+	}
+	n := it.it.NextMany(tmp[:min(len(dst), len(tmp))])
```
eguguchkin, line 273: I don't like that here — and in other implementations — tmp affects the size of dst. That makes responsibility blurry. Let's call NextMany multiple times if tmp is smaller than dst.

forshev (approved):

config/config.go
```diff
@@ -176,6 +176,15 @@ type Config struct {
 		} `config:"aggregation"`
 	} `config:"limits"`
 
+	QueryOptimization struct {
+		BatchExecution struct {
+			Enabled bool `config:"enabled"`
+			// CostThreshold is the minimum estimated non-batched execution cost required to enable batch-at-a-time query
+			// evaluation. Suggestion is to use value which is greater than 3 x LID block size.
+			CostThreshold int `config:"cost_threshold" default:"50000"`
```
dkharms: Default 50000 contradicts the comment right above:
> 3 x LID block size

node/util.go
```diff
@@ -0,0 +1,81 @@
...
+func NewBatcherNode(source Node, desc bool) BatchedNode {
+	return &batcherNode{
+		source: source,
+		desc:   desc,
+		batch:  make([]uint32, 0, maxBatchDrain),
+	}
+}
+
+func (b *batcherNode) NextBatch() LIDBatch {
+	batch := b.batch[:0]
```
dkharms: This is a little bit error-prone because at first I thought that `batcherNode` can wrap any other node, not only the root node (incorrect usage will lead to corruptions).

Maybe we should rename this type to specify that this type must wrap the root?

node/node.go
```diff
@@ -14,9 +14,9 @@ type Node interface {
 type BatchedNode interface {
 	fmt.Stringer
 	// NextBatch returns next batch. Returns nil when exhausted.
```
dkharms: It returns `emptyBatch`, not nil. Same for `NextBatchGeq` comment.

frac/processor/search.go
```diff
@@ -249,61 +247,86 @@ func iterateEvalTree(
...
+			n := iter.CopyLIDs(lidsBuf[:min(remaining, len(lidsBuf))], tmpBuf[:min(remaining, len(tmpBuf))])
+			timerEval.Stop()
+
+			if n == 0 {
+				break
+			}
+
+			lidsBatch := lidsBuf[:n]
+			total += n
+			remaining -= n
```
dkharms: We can silently drop some batches when lids that are mapped to the same id are returned.

Here we substract the lids count but [here](https://github.com/ozontech/seq-db/blob/d2eef53338328ad9262dd6c2294b7a686163e325/frac/processor/search.go#L314-L320) we perform deduplication.

Here is the small reproduction (it's LLM written but it is straight to the point):
```go
package processor

import (
	"context"
	"testing"

	"github.com/stretchr/testify/require"

	"github.com/ozontech/seq-db/metric/stopwatch"
	"github.com/ozontech/seq-db/node"
	"github.com/ozontech/seq-db/seq"
)

type stubIDsIndex struct {
	ids []seq.ID
}

func (s *stubIDsIndex) LessOrEqual(lid seq.LID, id seq.ID) bool { panic("not used") }
func (s *stubIDsIndex) GetMID(lid seq.LID) seq.MID              { return s.ids[lid].MID }
func (s *stubIDsIndex) GetRID(lid seq.LID) seq.RID              { return s.ids[lid].RID }
func (s *stubIDsIndex) Len() int                                { return len(s.ids) }

func (s *stubIDsIndex) GetMIDs(lids []node.LID, out []seq.MID) []seq.MID {
	for _, lid := range lids {
		out = append(out, s.ids[lid.Unpack()].MID)
	}
	return out
}

func (s *stubIDsIndex) GetRIDs(lids []node.LID, out []seq.RID) []seq.RID {
	for _, lid := range lids {
		out = append(out, s.ids[lid.Unpack()].RID)
	}
	return out
}

func TestIterateEvalTreeDuplicateIDsAtBatchBoundary(t *testing.T) {
	// LID 0 is unused; IDs descend as LID grows; LIDs 1 and 2 share one seq.ID.
	idx := &stubIDsIndex{ids: []seq.ID{
		1: {MID: 600, RID: 1},
		2: {MID: 600, RID: 1}, // duplicate of LID 1
		3: {MID: 500, RID: 1},
		4: {MID: 400, RID: 1},
		5: {MID: 300, RID: 1},
		6: {MID: 200, RID: 1},
	}}

	// Single batch with 6 LIDs, longer than the limit.
	evalTree := node.NewStaticBatched([]uint32{1, 2, 3, 4, 5, 6}, seq.DocsOrderDesc.IsReverse())

	params := SearchParams{
		Limit: 3,
		Order: seq.DocsOrderDesc,
	}

	total, ids, _, _, err := iterateEvalTree(context.Background(), params, idx, evalTree, nil, stopwatch.New())
	require.NoError(t, err)

	// 6 LIDs exist and the limit is not yet satisfied after the duplicate,
	// so the search must keep scanning: expected IDs are 600, 500, 400.
	got := make([]seq.MID, 0, len(ids))
	for _, id := range ids {
		got = append(got, id.ID.MID)
	}
	require.Equal(t, []seq.MID{600, 500, 400}, got)
	require.Equal(t, 4, total) // 4 LIDs scanned to produce 3 distinct IDs
}
```

dkharms (approved):

node/node_static.go
```diff
@@ -99,3 +99,91 @@ func MakeStaticNodes(data [][]uint32) []Node {
...
+	for {
+		if n.batch.IsEmpty() {
+			if n.ptr >= len(n.data) {
+				return EmptyBatch()
+			}
+			n.batch = NewSliceBatch(n.data[n.ptr:])
+			n.ptr = len(n.data)
+		}
+
+		if n.batch.IsEmpty() {
```
eguguchkin, line 145: Looks like an unnecessary check here

node/node_static.go
```diff
@@ -99,3 +99,91 @@ func MakeStaticNodes(data [][]uint32) []Node {
...
+	for {
+		if n.batch.IsEmpty() {
+			if n.ptr < 0 {
+				return EmptyBatch()
+			}
+			n.batch = NewSliceBatch(n.data[:n.ptr+1])
+			n.ptr = -1
+		}
+
+		if n.batch.IsEmpty() {
```
eguguchkin, line 177: Looks like an unnecessary check here

node/util.go
```diff
@@ -0,0 +1,81 @@
...
+
+func NewBatcherNode(source Node, desc bool) BatchedNode {
+	return &batcherNode{
+		source: source,
+		desc:   desc,
+		batch:  make([]uint32, 0, maxBatchDrain),
+	}
+}
+
+func (b *batcherNode) NextBatch() LIDBatch {
```
eguguchkin, line 27: Looks unnecessarily complicated.

Here's the same thing without local batch and pooled.


```go
func (b *batcherNode) NextBatch() LIDBatch {
	b.batch = b.batch[:0]
	for len(b.batch) < maxBatchDrain {
		lid := b.source.Next()
		if lid.IsNull() {
			break
		}
		b.batch = append(b.batch, lid.Unpack())
	}
	if !b.desc {
		slices.Reverse(b.batch)
	}
	return NewSliceBatch(b.batch)
}

```

node/util.go
```diff
@@ -0,0 +1,81 @@
...
+	b.batch = batch[:0]
+	if !b.desc {
+		slices.Reverse(batch)
+	}
+	return NewSliceBatch(batch)
+}
+
+func (b *batcherNode) NextBatchGeq(nextID LID) LIDBatch {
+	batch := b.batch[:0]
+	polled := 0
```
eguguchkin: Same here — batch and pooled don't really make sense in this context

frac/processor/batch_eval_tree.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+		childCosts := make([]uint64, len(root.Children))
+		for i, child := range root.Children {
+			c, err := calculateQueryIterationCost(child, ti, cache)
+			if err != nil {
+				return 0, err
+			}
+			childCosts[i] = c
+		}
+		if len(childCosts) != 2 {
```
eguguchkin, line 111: nit: Looks like a second check for 'not'. Is it definitely needed here?

frac/processor/batch_eval_tree.go
```diff
@@ -0,0 +1,223 @@
...
+		return 0, fmt.Errorf("unsupported token type for cost estimation")
+	}
+}
+
+func leafIterationCost(ti tokenIndex, field string, token parser.Token, cache leafTIDsCache) (uint64, error) {
+	tids, err := ti.GetTIDsByTokenExpr(token)
+	if err != nil {
+		return 0, err
+	}
+	if len(tids) > maxBatchedTIDsPerLeaf {
```
eguguchkin, line 136: I just don't fully understand the criteria for using batches here

frac/processor/batch_eval_tree.go
```diff
@@ -0,0 +1,223 @@
...
+		return 0, errBatchingUnsupported
+	}
+	cache[token] = tids
+	if len(tids) == 0 {
+		return 0, nil
+	}
+
+	freqs := ti.GetFreqsByTIDs(tids, field)
```
eguguchkin, line 144: What do you think — could this method be merged with GetTIDsByTokenExpr in the future to extract data in a single pass?

frac/processor/batch_eval_tree.go
```diff
@@ -0,0 +1,223 @@
...
+) (node.BatchedNode, error) {
+	if !queryOpts.BatchExecution.Enabled {
+		return nil, errBatchingUnsupported
+	}
+
+	if !astSupportsBatching(root) {
+		return nil, errBatchingUnsupported
+	}
+
+	cache := make(leafTIDsCache)
```
eguguchkin, line 46: Just thinking out loud — maybe in the future we could create an intermediate structure for the query execution plan, like AST → PlanTree, and this populated plan would contain all the TIDs, costs, and logical nodes, which would then be used to build the LIDs processing tree. That way we wouldn't have to make the same index lookups (GetTIDsByTokenExpr) repeatedly through the cache.

eguguchkin (approved):

frac/processor/batch_eval_tree.go
cheb0: yes

frac/processor/batch_eval_tree.go
cheb0: The whole point it to actually avoid (for now) queries like 'service:abc* AND level:[0 to 7]'. Additional work will be needed to support them.

config/config.go
cheb0: set default value to 150000

node/node.go
cheb0: fixed

PR #490, opened 2026-08-11
perf: galloping search in sealed LID iterators' NextGeq

radmirnovii:
# Description

`NextGeq` in the sealed LID iterators runs a binary search over the remaining
chunk slice on every call. The AND/OR zigzag sends monotonically increasing
targets while the slice narrows from the consumption edge, so the answer is
almost always near the edge — instrumented on a 2M-doc fraction: ~50% of calls
land at distance 0 and 99.9% within 16, while binary search pays
log2(remainder) ≈ 14 probes regardless. This PR replaces it with an
exponential (galloping) search from the edge, ~2·log2(d) probes — the same
uncapped scheme as Lucene's
[`IntArrayDocIdSet.advance`](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java)
and RoaringBitmap's
[`advanceUntil`](https://github.com/RoaringBitmap/roaring/blob/master/setutil.go),
which seq-db already depends on for skip masks.

The worst case is bounded at ~2× binary (2·log2(d) ≤ 2·log2(n)) and far jumps
are self-limiting (jumps over one chunk sum to at most its length): an AND
built deliberately from a skewed pair (freq ratio ≈1450, 71% of calls in the
far zone) measures at parity. A capped variant was measured and rejected — it
hurts the common zone and rescues regimes that did not occur across ~6M
instrumented calls (max distance < 128).

Benchstat, n=10, in-process `Sealed.Search`, 2M-doc fraction, vs `main`:

| scenario | diff |
|---|---|
| 3-token AND | **−21.8%** |
| `(a OR b) AND c` | **−10.8%** |
| 2-token AND | **−4.5%** |
| skewed ANDs (ratio ≈285 / ≈1450) | ~ |
| 13 other scenarios (needle hit/miss, aggregations, histogram, wildcard/contains, fetch, active fraction, pure OR, AND NOT) | ~ |

Search results are byte-identical to `main` on a 57-query corpus (IDs and
order in both sort directions, aggregation bins, fetched docs).

Note: `fracmanager/TestCapacityExceeded` is red on current `main` (a73114c)
independently of this change.

---

- [x] I used LLM/AI assistance to make this pull request;

```
Model: Claude Fable 5 (claude-fable-5), via Claude Code
Prompt: interactive session, no single prompt. Task: performance research of
seq-db search internals (profiling, microbenchmarks, adopt/reject verdicts
backed by numbers). Directives shaping this PR: deep-dive galloping; prove
edge-locality with in-engine instrumentation rather than synthetic tests;
write a standalone PR with edge-case/regression justification and prior-art links
Extent: code (gallop.go, test, call sites), benchmarks, and this PR text
were AI-generated under the author's direction and review.
```

### DIFF
--- a/frac/sealed/lids/gallop.go
+++ b/frac/sealed/lids/gallop.go
@@ -1,0 +1,66 @@
+package lids
+
+// Galloping (exponential) searches for the NextGeq hot path: zigzag targets
+// are monotone and the chunk slice narrows from the consumption edge, so the
+// answer is usually a few elements away from the edge. Probing exponentially
+// from the edge costs ~2*log2(d) for an answer d away (bounded by ~2x a full
+// binary search over the remainder, which pays log2(len) regardless of d).
+
+// searchGeqGallop returns the index of the first element >= t, probing from
+// the front (IteratorDesc consumes the slice front-to-back).
+func searchGeqGallop(a []uint32, t uint32) int {
+	n := len(a)
+	if n == 0 || a[0] >= t {
+		return 0
+	}
+	i := 1
+	for i < n && a[i] < t {
+		i <<= 1
+	}
+	// a[i>>1] < t, so the answer is in (i>>1, min(i, n)].
+	lo := i>>1 + 1
+	hi := i
+	if hi > n {
+		hi = n
+	}
+	for lo < hi {
+		m := int(uint(lo+hi) >> 1)
+		if a[m] < t {
+			lo = m + 1
+		} else {
+			hi = m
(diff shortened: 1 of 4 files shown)


eguguchkin:
@seqbenchbot up main

eguguchkin:
@seqbenchbot --help

eguguchkin:
@seqbenchbot start search-aggregation

PR #496, opened 2026-08-13
feat(StreamSearch): add stream search on stores

### LINKED ISSUE #483: Stream search on stores
StreamSearch method now uses stream search implementation based on https://github.com/ozontech/seq-db/pull/360

forshev:
# Description

Add StreamSearch method on stores
Add Executors
Serve StreamSearch with executors

Fixes #483

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```


query/record.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+// Executors make use of val's index. the plan knows which executors use which col indexes
+type RecordVals struct {
+	Type DataType
+
+	// for lazy decoding
+	rawData []byte // raw data
+
+	decoded any
```
eguguchkin, line 45: **todo**: From a performance perspective, any here is not a cache but just another container. We'll need to reconsider the containerization approach in the next iteration.

query/exec/filter.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+}
+
+func (e *Lt[T]) Eval(other T) bool {
+	return other < e.pred
+}
+
+type DocFilter struct {
+	field  string
+	filter FilterExpr[string]
```
eguguchkin, line 142: **todo**: I'd like to propose a different approach: instead of having filters work with documents directly, we could introduce an extractor that retrieves scalar values from JSON. Then we'd combine the filter pipeline with the extraction pipeline. The reasoning is that if we have multiple filters, parsing JSON for each filter separately would be costly. A shared extractor would parse once and extract all the values we might need in the future.

query/exec/datasource.go
```diff
@@ -0,0 +1,293 @@
...
+		}
+		if s.done {
+			return nil
+		}
+	}
+
+	idSrc := s.qpr.IDs[s.curIdx]
+	s.curIdx++
+
+	docs, err := s.fetcher.FetchDocs(s.Ctx(), s.fracs, []seq.IDSource{idSrc}, false)
```
eguguchkin: It seems we can pass noSkipMask = true here.

query/exec/datasource.go
```diff
@@ -0,0 +1,293 @@
...
+
+	if s.fracs == nil {
+		s.fracs, s.release = s.fracManager.AcquireFractionsInRange(params.From, params.To)
+	}
+
+	params.Limit = searcherBatchLimit
+
+	if s.batchNo > 0 {
+		// Advance the cursor and narrow the time range so the fractions filter skips already-served data.
+		params.WithTotal = false
```
eguguchkin, line 167: **todo**: I think we should reconsider the withTotal logic. My suggestion is to apply withTotal to the last batch, or ideally to finalize. The reason is that our datasource is infinite and streams everything it finds. Totals should only be calculated when an executor imposes a limit and triggers finalize. That would be the right moment to compute all the rest totals and include them in the summary.

query/exec/datasource.go
```diff
@@ -0,0 +1,293 @@
...
+		lastID := s.qpr.IDs[len(s.qpr.IDs)-1].ID
+		params.OffsetId = lastID
+		if params.Order == seq.DocsOrderDesc {
```
eguguchkin: I think this is already done inside getLIDsBorders. Could we double-check whether we really need to align from and to here as well?

query/exec/aggregator.go
```diff
@@ -0,0 +1,188 @@
...
+		if s == nil {
+			continue
+		}
+		total += s.Total
+		if firstErr == nil && s.Err != nil {
+			firstErr = s.Err
+		}
+	}
+	if firstErr == nil {
+		firstErr = a.err
```
eguguchkin: a.err appears to be never assigned — it always remains nil/default. Should we remove it, or is there a missing assignment somewhere?

query/exec/datasource.go
```diff
@@ -0,0 +1,293 @@
...
+
+	searchParams processor.SearchParams
+	isAgg        bool
+
+	fracManager *fracmanager.FracManager
+	searcher    *fracmanager.Searcher
+	fetcher     *fracmanager.Fetcher
+
+	qpr         *seq.QPR
+	aggs        []*storeapi.SearchResponse_Agg
```
eguguchkin: I think we should also remove the multiple aggregation here and always use a single aggregation instead of a slice.

query/exec/merger.go
```diff
@@ -0,0 +1,281 @@
...
+		return r
+	}
+
+	if m.curRight == nil {
+		r := m.curLeft
+		m.curLeft = m.left.Next()
+		return r
+	}
+
+	leftVal := m.extractValue(m.curLeft)
```
eguguchkin, line 116: **todo**: Similar to what I suggested for the filter — we should consider extracting all required values from the document once, and then pass them as scalars within one of the columns. This way we avoid parsing the JSON repeatedly at every stage, which would be inefficient.

query/exec/merger.go
```diff
@@ -0,0 +1,281 @@
...
+	less := createLessFunc()
+
+	return &Merger{
+		left:     left,
+		right:    right,
+		colIdx:   colIdx,
+		field:    field,
+		dataType: dataType,
+		order:    order,
+		less:     less,
```
eguguchkin: less -> cmp

query/exec/merger.go
```diff
@@ -0,0 +1,281 @@
...
+) query.RecordProducer {
+	if len(producers) == 0 {
+		return &emptyRecordProducer{}
+	}
+
+	if len(producers) == 1 {
+		return NewMerger(producers[0], &emptyRecordProducer{}, colIdx, field, dataType, order)
+	}
+
+	merged := NewMerger(producers[0], producers[1], colIdx, field, dataType, order)
```
eguguchkin, line 266: **todo:** We should implement a balanced tree here to ensure consistent performance

query/exec/merger.go
```diff
@@ -0,0 +1,281 @@
...
+	case query.DataTypeInt64:
+		return decoded.(int64)
+	case query.DataTypeFloat64:
+		return decoded.(float64)
+	default:
+		return ""
+	}
+}
+
+func createLessFunc() func(any, any) int {
```
eguguchkin: **todo**: Doing a type check for each comparison is costly. We should determine the type once based on m.dataType and return the corresponding comparison function, rather than checking it repeatedly.

query/exec/projector.go
```diff
@@ -0,0 +1,97 @@
...
+
+	decoder := r.Vals[p.colIdx].Decoded().(*insaneJSON.Root)
+
+	var newRecord *query.Record
+	if !p.filter.AllowList {
+		// It is block list, so remove given fields from document.
+		for _, field := range p.filter.Fields {
+			decoder.Dig(field).Suicide()
+		}
+		newRecord = p.makeRecordWithNewVals(r, decoder.Encode(p.decoderBuf[:0]))
```
eguguchkin: We're kind of reusing a buffer here, but fortunately it doesn't actually get reused because it starts empty, and inside Encode() a new slice is created via Grow. Otherwise, we'd end up with a dirty slice. To avoid confusion, let's just remove the buffer altogether and call Encode(nil) — that's effectively what's happening anyway.

query/exec/projector.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+		for _, field := range fieldsToRemove {
+			field.Suicide()
+		}
+		newRecord = p.makeRecordWithNewVals(r, decoder.Encode(p.decoderBuf[:0]))
+	}
+
+	// The input record holds the mutated root and won't be seen downstream;
+	// keep it for release in Finalize.
+	p.roots = append(p.roots, r)
```
eguguchkin: I don't see the point of accumulating them until finalize — we don't reuse them afterwards.

storeapi/grpc_stream_search.go
```diff
@@ -0,0 +1,477 @@
...
+	if err != nil {
+		return fmt.Errorf("store can't stream records: %w", err)
+	}
+
+	// CANCEL: terminate immediately, no summary.
+	if outcome == outcomeCancel {
+		return nil
+	}
+
+	metric.SearchDurationSeconds.Observe(time.Since(start).Seconds())
```
eguguchkin: лучше выделить отдельную метрик потому что значения сравнивать со старым методом не корректно - там нет стадии фетча

storeapi/grpc_stream_search.go
```diff
@@ -0,0 +1,477 @@
...
+
+	summary := producer.Finalize()
```
eguguchkin, line 184: **todo**: In general, I'd like to discuss a couple of things again:

- Context cancellation on the client side vs ControlAction_CANCEL — does one replace the other?
- Premature stop via ControlAction_FINALIZE vs reading until the end — what should summary.Total contain in the first case? It seems like the total would be invalid for this scenario.

storeapi/grpc_stream_search.go
```diff
@@ -0,0 +1,477 @@
...
+	}()
+
+	outcome, err := g.streamSearchRecords(ctx, stream, producer, controlCh, recvErrCh)
+	if err != nil {
+		return fmt.Errorf("store can't stream records: %w", err)
+	}
+
+	// CANCEL: terminate immediately, no summary.
+	if outcome == outcomeCancel {
+		return nil
```
eguguchkin, line 179: It seems that producer.Finalize() still needs to be called (just not sent) because we have resources there that need to be released.

Right, otherwise we'll never release the fractions!

And Finalize must be called in any case, regardless of errors.

storeapi/grpc_stream_search.go
```diff
@@ -0,0 +1,477 @@
...
+
+	err = stream.Send(&storeapi.StreamSearchResponse{
+		ResponseType: &storeapi.StreamSearchResponse_Header{
+			Header: &storeapi.ResponseHeader{
+				Typing: typing,
+			},
+		},
+	})
+	if err != nil {
+		if util.IsCancelled(ctx) {
```
eguguchkin, line 136: I think we should call _ = producer.Finalize() here too. While the current implementation doesn't allocate any resources that need releasing, the interface contract might evolve, and future implementations could require cleanup. We should be defensive and ensure Finalize is called. Let's also audit all other return points to see if Finalize is needed there as well.

proxy/search/stream_search.go
```diff
@@ -0,0 +1,452 @@
...
+	err := it.stream.CloseSend()
+	it.tr.Done()
+	return err
+}
+
+func (it *StreamSearchIterator) Finalize() *query.Summary {
+	// If the stream was finalized before the data was exhausted, the store's summary may still be in flight.
+	// Drain the remaining messages so the store-reported summary is not lost.
+	if !it.done {
+		it.drain()
```
eguguchkin, line 381: We need to call finalize before drain to avoid doing unnecessary work on the store

proxyapi/grpc_stream_search.go
```diff
@@ -48,28 +52,36 @@ func (g *grpcV1) StreamSearch(stream seqproxyapi.SeqProxyApi_StreamSearchServer)
...
 	if err != nil {
-		return err
-	}
-	if sResp.err != nil && sResp.err.Code == seqproxyapi.ErrorCode_ERROR_CODE_PARTIAL_RESPONSE && shouldFailPartialResponse(ctx) {
-		return status.Error(codes.Internal, "partial response: not all shards returned results")
-	}
-	if sResp.err != nil && !shouldHaveResponse(sResp.err.Code) {
-		return errors.New(sResp.err.Message)
+		// The stores were not opened or failed to open, cancel any that may have started before propagating the error.
+		if broadcaster != nil {
```
eguguchkin: I'm not sure I understand why we always send cancel here. Also, in the case of a partial response, we seem to continue trying to stream even after cancellation — is that intended?

query/exec/datasource.go
```diff
@@ -0,0 +1,293 @@
...
+// full total once); subsequent batches drop WithTotal and advance the cursor
+// (OffsetId) from the previous batch's last ID, narrowing the time range.
+func (s *SearcherDataSource) scanBatch() error {
+	params := s.searchParams
+
+	if s.fracs == nil {
+		s.fracs, s.release = s.fracManager.AcquireFractionsInRange(params.From, params.To)
+	}
+
+	params.Limit = searcherBatchLimit
```
eguguchkin: I believe we should use the original limit rather than hardcoding 1000. Usually, a page request asks for 100 documents — fetching 1000 at a time seems unnecessary.

eguguchkin (approved):

proxy/search/stream_search.go
forshev: we already send finalize to stores right before calling finalize on producer. it just happens outside of producer

proxyapi/grpc_stream_search.go
forshev: fixed

query/exec/datasource.go
forshev: fixed

storeapi/grpc_stream_search.go
forshev: fixed

storeapi/grpc_stream_search.go
forshev: fixed

query/exec/datasource.go
forshev: fixed

query/exec/datasource.go
forshev: fixed

query/exec/aggregator.go
forshev: fixed

query/exec/datasource.go
forshev: fixed

query/exec/merger.go
forshev: fixed

query/exec/projector.go
forshev: fixed

query/exec/projector.go
forshev: fixed

proxy/search/stream_search.go
eguguchkin: **todo:**

- All calls to broadcaster.SendControl(storeapi.ControlAction_FINALIZE) are made right before summary := storesStream.Finalize(). It seems like the broadcaster isn't needed here as a FINALIZE sender — this could be hidden inside storesStream.Finalize().
- As for broadcaster.SendControl(storeapi.ControlAction_CANCEL) — we need to figure out how to replace it or hide it behind a context cancel().

api/storeapi/store_api.proto
```diff
@@ -265,3 +267,83 @@ message StatusRequest {}
...
+message StreamSearchResponse {
+  oneof ResponseType {
+    ResponseHeader header = 1;
+    ResponseData data = 2;
+    ResponseSummary summary = 3;
+  }
+}
+
+message ResponseHeader {
+  repeated Typing typing = 1;
```
dkharms, line 306: Do I understand correctly that `ResponseHeader` (and therefore type information) describes all values that are returned from store as `Record`?

So, for example, if result of aggregation, like (I am using PRQL syntax here):
```
filter foo == "bar"
group {service, handler} (
  aggregate {avg_duration = average duration}
)
```

header will contain `[(avg_duration, FLOAT64); (service, STRING); (handler, STRING)]`? And each record must satisfy this constraint, right?

api/storeapi/store_api.proto
```diff
@@ -265,3 +267,83 @@ message StatusRequest {}
...
+  bool explain = 4;                   // Should request be explained (tracing will be provided with the result).
+  string offset_id = 5;               // ID offset for pagination.
+  bool with_total = 6;                // Should total number of documents be returned in response.
+}
+
+message StreamControl {
+  ControlAction action = 1;
+}
+
+enum ControlAction {
```
dkharms, line 291: IIRC, we were thinking about pull-based model where seqproxy will pull data from store in blocks so it could have more control on memory usage. Should we add `PULL` control action here as well? Or we rely on gRPC protocol and `stream.Recv`?

P.S. Sorry, I am just starting reviewing this PR so I may ask stupid questions.

api/storeapi/store_api.proto
```diff
@@ -265,3 +267,83 @@ message StatusRequest {}
...
+    ResponseData data = 2;
+    ResponseSummary summary = 3;
+  }
+}
+
+message ResponseHeader {
+  repeated Typing typing = 1;
+}
+
+message Typing {
```
dkharms, line 309: Also I believe we should add (maybe even here) encoding we've used for this type (or block of records). What do you think? Of course, it is out of scope of this PR.

api/storeapi/store_api.proto
```diff
@@ -265,3 +267,83 @@ message StatusRequest {}
...
+message ResponseData {
+  RecordsBatch batch = 1;
+}
+
+message RecordsBatch {
+  repeated Record records = 1;
+}
+
+message Record {
+  repeated bytes raw_data = 1;
```
dkharms, line 337: Why `Record` consists of array of byte arrays? Why not just single byte array?

query/exec/merger.go
dkharms: @forshev you've resolved the comment but the tree here is still not balanced. This will be done in other PR?

parser/seqql_pipes.go
```diff
@@ -101,6 +110,62 @@ func parsePipes(lex *lexer) ([]Pipe, error) {
...
+}
+
+func (f *PipeFilter) DumpSeqQL(o *strings.Builder) {
+	o.WriteString("filter ")
+	o.WriteString(quoteTokenIfNeeded(f.Condition.Field))
+	o.WriteString(":")
+	o.WriteString(quoteTokenIfNeeded(f.Condition.Value))
+}
+
+func parsePipeFilter(lex *lexer) (*PipeFilter, error) {
```
dkharms, line 133: It is worth adding examples of correct pipes usage for each pipe.

query/exec/merger.go
```diff
@@ -0,0 +1,281 @@
...
+	// once chosen, so this is the last owner and Finalize releases them. On the
+	// documents path the merger compares by SeqID and the vals stay undecoded,
+	// so roots stays empty; the DataTypeDocument path is handled for
+	// correctness.
+	roots []*query.Record
+
+	done bool
+}
+
+func NewMerger(
```
dkharms, line 42: I do not think that this should be part of public API -- `NewNMergedProducers` is enough.

query/exec/merger.go
```diff
@@ -0,0 +1,281 @@
...
+			return cmp.Compare(v, b.(float64))
+		case string:
+			return cmp.Compare(v, b.(string))
+		default:
+			return 0
+		}
+	}
+}
+
+func NewNMergedProducers(
```
dkharms, line 251: Why not `NewDistributedMerger`?

query/exec/filter.go
```diff
@@ -0,0 +1,158 @@
+package exec
+
+import (
+	"cmp"
+
+	insaneJSON "github.com/ozontech/insane-json"
```
dkharms, line 6: I think usage of `insane-json` for now is just a overcomplication -- it brings to the table a lot of unnecessary memory-management.

query/exec/datasource.go
dkharms: In what case `withTotal` will be required in the future? In my opinion this must be expressed via query language.

So if one wants to receive 1k documents and total number of documents he will have to write two different queries:
```
foo:bar | limit 1000  # Get the first thousand of documents
foo:bar | count       # Get the overall amount of such documents
```

query/record.go
```diff
@@ -0,0 +1,118 @@
...
+
+const (
+	DataTypeBytes DataType = iota
+	DataTypeSeqID
+	DataTypeDocument
+	DataTypeString
+	DataTypeUint32
+	DataTypeUint64
+	DataTypeInt32
+	DataTypeInt64
```
dkharms, line 32: Maybe we can reduce to `UINT`, `INT` and `FLOAT`? And make them 64 bits by default.

dkharms (approved):
In overall I like the idea. Would be interesting to see how it will evolve :)

api/storeapi/store_api.proto
forshev: each inner array represents single "field" of a "table". it's easier to pack and unpack this way
and we get partial unpacking of a record out-of-the-box

api/storeapi/store_api.proto
forshev: yes, `stream.Recv` acts as a pull command

api/storeapi/store_api.proto
forshev: this is correct

query/exec/merger.go
forshev: yes, all the comments that start with `todo` will be addressed later

parser/seqql_pipes.go
forshev: we have them in tests

query/exec/datasource.go
forshev: withTotal is one of those moments that we need to discuss

query/record.go
forshev: it's possible

query/exec/filter.go
forshev: maybe we'll replace insane-json with something else

api/storeapi/store_api.proto
forshev: we will need it if there will be multiple ways to encode the same data, which is possible

query/exec/merger.go
eguguchkin: We might want to move each individual fraction to a producer (when we implement timestamp binning) and merge them into a single stream at the store using the same executor.

PR #499, opened 2026-08-18
refactor: fair hit-rate for registry

dkharms:
# Description

In this PR I've reintroduced fix for too high registry cache hit-rate.

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/frac/sealed.go
+++ b/frac/sealed.go
@@ -40,9 +40,8 @@ type Sealed struct {
 	docsReader storage.DocsReader
 
 	// IsLegacy is true for fractions that use the old single .index file format.
-	IsLegacy     bool
-	legacyFile   *os.File
-	legacyReader storage.IndexReader
+	IsLegacy   bool
+	legacyFile *os.File
 
 	// Per-section index files and their readers (new split format only).
 	infoFile    *os.File
@@ -163,12 +162,6 @@ func (f *Sealed) openIndex() {
 		// We have exactly one `.index` file for legacy sealed fractions.
 		// So opening only this file is sufficient.
 		f.legacyFile = f.openFileIfNeeded(f.legacyFile, f.BaseFileName+consts.IndexFileSuffix)
-
-		f.legacyReader = storage.NewIndexReader(
-			f.readLimiter, f.legacyFile.Name(),
-			f.legacyFile, f.indexCache.LegacyRegistry,
-		)
-
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


frac/sealed.go
```diff
@@ -544,41 +491,36 @@ func (f *Sealed) IsIntersecting(from, to seq.MID) bool {
...
+		LID: storage.NewIndexReader(
+			f.readLimiter, f.lidFile.Name(), f.lidFile,
+			cache.NewSession(f.indexCache.LIDRegistry),
+		),
 	}
-
-	return bi.Info, nil
 }
 
 // computeIndexOnDisk returns the total on-disk size of index files for a local fraction.
```
forshev: nit: comment doesn't match method's title

forshev (dismissed):

frac/sealed.go
```diff
@@ -160,94 +158,42 @@ func NewSealedPreloaded(
...
+		f.legacyFile = f.openFileIfNeeded(f.legacyFile, f.BaseFileName+consts.IndexFileSuffix)
 
-	name := f.BaseFileName + consts.IndexFileSuffix
-	file, err := os.Open(name)
-	if err != nil {
-		logger.Fatal(
-			"can't open legacy index file",
-			zap.String("file", name),
-			zap.Error(err),
+		f.legacyReader = storage.NewIndexReader(
```
eguguchkin: nit: Here the reader is not used, so there’s no need to initialise it. Even if it’s used somewhere later, it’s better to move the initialisation to that place.

I’d also remove it entirely from the permanent properties of Sealed.legacyReader.

frac/sealed.go
```diff
@@ -283,16 +229,30 @@ func (f *Sealed) loadInfo() {
 	var err error
 
 	if f.IsLegacy {
-		f.openInfoLegacy()
+		f.legacyFile = f.openFileIfNeeded(f.legacyFile, f.BaseFileName+consts.IndexFileSuffix)
+
+		f.legacyReader = storage.NewIndexReader(
```
eguguchkin: nit: I understand this is a one-time load, but for consistency, I would also create a reader via Session here. 

I'd make it local (not a property of Sealed). 

I also suggest doing the same for remote.

eguguchkin (approved):

forshev (approved):

PR #503, opened 2026-08-19
feat(stream search): serve regular methods through stream search

### LINKED ISSUE #498: Serve ComplexSearch request through stream search if approrpiate
if:

- config flag is enabled
- search request is suitable for stream search (no hist, only one agg, etc)

Serve ComplexSearch request through stream (proxy <-> store) search

forshev:
# Description

- ComplexSearch and GetAggregation methods try to serve requests through stream search if request is suitable (no histogram, only one aggregation) and config flag is enabled.
- Support quantile and unique count aggregation functions in stream search 

Fixes #498

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```


config/config.go
```diff
@@ -354,6 +354,8 @@ type Config struct {
 		// Specify how many tokens can be checked using regular expressions.
 		// If zero then there is no limit.
 		MaxRegexTokensCheck int `config:"max_regex_tokens_check" default:"0"`
+		// If true, suitable ComplexSearch queries will be served through stream search implementation.
+		TryStreamSearch bool `config:"try_stream_search"`
```
cheb0: I think the name is a bit ambigous

proxyapi/grpc_complex_search.go
```diff
@@ -74,3 +90,200 @@ func aggregationArgsFromProto(aggs []*seqproxyapi.AggQuery) []seq.AggregateArgs
...
+	summary := storesStream.Finalize()
+	if summary == nil {
+		summary = &query.Summary{}
+	}
+	if partialErr != nil && summary.Err == nil {
+		summary.Err = partialErr
+	}
+
+	if req.Query.Explain {
+		tr.Done()
```
cheb0: nit: in `ComplexSearch` `tr.Done()` is always called, not matter the `explain` field

parser/seqql.go
```diff
@@ -74,6 +75,37 @@ func parse(q string, mapping seq.Mapping) (SeqQLQuery, error) {
 	}, nil
 }
 
+// trimOuterParens removes every layer of parentheses that wraps the entire
+// query, e.g. "(((* | fields level)))" -> "* | fields level". Parentheses that
+// only wrap part of the query (e.g. "(a:1) | fields b") are left untouched.
+func trimOuterParens(q string) string {
```
cheb0, line 81: Can you also write a comment on why we are doing this?

proxyapi/grpc_complex_search.go
```diff
@@ -74,3 +90,202 @@ func aggregationArgsFromProto(aggs []*seqproxyapi.AggQuery) []seq.AggregateArgs
 	}
 	return args
 }
+
+func (g *grpcV1) emulateStreamSearch(
```
cheb0: I think the name is a bit odd. We actually emulate complex search using stream search, but the name states the opposite.

config/config.go
cheb0: Let's say we enabled `try_stream_search` on prod. Now every aggregation request with one agg will go to stream search. My question - do we have same correctness guarantees like we have for an ordinary aggregation? For instance ordinary agg is covered using `fraction_test` and aggs are tested on both low and high cardinality fields.

cheb0 (approved):

parser/seqql.go
forshev: frontend now wraps every query with parentheses, they use it for predefined filters like "only with traces".
it breaks queries with pipes. we will fix the frontend, but i thought it would be useful to handle this case in seqdb aswell

proxyapi/grpc_complex_search.go
forshev: renamed to `useStreamSearch`

config/config.go
forshev: for now we have only basic integration tests and unit tests for executors

proxyapi/grpc_complex_search.go
forshev: fixed

parser/seqql.go
```diff
@@ -74,6 +75,37 @@ func parse(q string, mapping seq.Mapping) (SeqQLQuery, error) {
...
+				if depth == 0 && i != len(q)-1 {
+					outer = false
+				}
+			}
+		}
+		if !outer {
+			return q
+		}
+
+		q = strings.TrimSpace(q[1 : len(q)-1])
```
eguguchkin: String literals are not taken into account, and they can also contain brackets. Add this to the test and it will fail.

Example test case:

```
test(`(((* | fields:"le)vel")))`, `* | fields:"le)vel"`)
```

query/exec/datasource.go
```diff
@@ -267,7 +276,16 @@ func makeDocumentRecord(id seq.ID, payload []byte) *query.Record {
...
 
-func makeAggRecord(bin *storeapi.SearchResponse_Bin) *query.Record {
+func makeAggRecord(bin *storeapi.SearchResponse_Bin, valuesPool []string) *query.Record {
+	// For unique_count, the store holds unique field values as indices into the per-agg ValuesPool.
+	// Resolve them to strings here so each record is self-contained, no shared pool.
+	var values []string
+	if len(bin.Hist.Values) > 0 && len(valuesPool) > 0 {
+		values = make([]string, 0, len(bin.Hist.Values))
+		for _, idx := range bin.Hist.Values {
+			values = append(values, valuesPool[idx])
```
eguguchkin, line 286: **todo**: This will be very expensive to pass like that, but it's not clear what would be better? We'll need to think about it separately — maybe in the header as typing? Or in the summary? And what if the list is very large? Should we split it into messages?

eguguchkin (approved):

parser/seqql.go
forshev: fixed

PR #512, opened 2026-09-05
fix(stream search): fix grpc connection leak

forshev:
# Description

Fixed an issue where connections weren't released in the StreamSearch method

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```

### DIFF
--- a/proxy/search/stream_search.go
+++ b/proxy/search/stream_search.go
@@ -369,7 +369,11 @@ func (it *StreamSearchIterator) SendControl(action storeapi.ControlAction) error
 // It is best-effort and safe to call on an already-closed stream; it must not be called concurrently with Next/Finalize.
 func (it *StreamSearchIterator) Close() error {
 	_ = it.SendControl(storeapi.ControlAction_CANCEL)
+	if !it.done {
+		it.drain()
+	}
 	err := it.stream.CloseSend()
+	it.recvTrailer()
 	it.tr.Done()
 	return err
 }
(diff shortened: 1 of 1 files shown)


proxy/search/stream_search.go
```diff
@@ -380,10 +384,31 @@ func (it *StreamSearchIterator) Finalize() *query.Summary {
...
+		case errors.Is(err, io.EOF):
+			return // trailer consumed, the stream is released
+		default:
+			if it.err == nil {
+				it.err = err
+			}
+			return
+		}
+	}
+}
```
dkharms, line 410: This is some kind of overcomplication. We ALWAYS have to drain inner buffers of gRPC stream. So why not include it in regular `drain()`?:
```go
func (it *StreamSearchIterator) Close() error {
	_ = it.SendControl(storeapi.ControlAction_CANCEL)
	it.drain()

	err := it.stream.CloseSend()
	it.tr.Done()

	return err
}

func (it *StreamSearchIterator) Finalize() *query.Summary {
	it.drain()

	_ = it.stream.CloseSend()
	it.tr.Done()

	return &query.Summary{Total: it.total, Err: it.err}
}

func (it *StreamSearchIterator) drain() {
	for {
		msg, err := it.stream.Recv()
		if err != nil {
			it.done = true
			if errors.Is(err, io.EOF) {
				return
			}
			it.err = err
			return
		}

		if err := it.push(msg); err != nil {
			it.err = err
		}
	}
}
```

By the way, `CANCEL` action serves no purpose -- stream cancellation is a part of gRPC protocol so I suggest to remove this option and stick only with `FINALIZE`.

eguguchkin: Agreed, it's overcomplication. I'd actually go even further with refactoring this logic. For instance, the done flag is inconsistent — it's set in multiple places, triggered both by EOF and after the first Summary in different parts of the incoming data processing flow.

My suggestion: let's implement the basic simplification from the example right now, and address the rest in future tasks.

eguguchkin (approved):

proxy/search/stream_search.go
forshev: will address this later

dkharms (approved):

PR #527, opened 2026-09-16
refactor(query engine): parallel inputs reading in aggregator

forshev:
# Description

Parallel inputs reading in aggregator executor

---

```
Model: {{model-name}}
Prompt: {{prompt}}
```


query/exec/aggregator.go
```diff
@@ -171,6 +129,59 @@ func (a *DistributedAggregator) Next() *query.Record {
...
+func (a *DistributedAggregator) drainInput(input query.RecordProducer) {
+	for {
+		r := input.Next()
+		if r == nil {
+			break
+		}
+
+		key := aggKey{
+			token: r.Vals[0].Decoded().(string),
+			ts:    r.Vals[6].Decoded().(uint64),
```
dkharms, line 141: We need to introduce some schema for these records -- now these constants are hard-coded all over the codebase.

Maybe we can do something like:
```go
type Col[T any] struct {
      Idx   int
      Type  DataType
}

func (r *Record) Get[T any](c Col[T]) T {
      return r.Vals[c.Idx].Decoded().(T)
}
```

I am fine with leaving `TODO` for that and creating an issue.

query/exec/aggregator.go
```diff
@@ -59,58 +62,13 @@ func NewDistributedAggregator(
 
 func (a *DistributedAggregator) Next() *query.Record {
 	if a.state == ExecutorStateReadingInput {
-		// TODO: read from all inputs simultaneously (???)
+		var wg sync.WaitGroup
```
dkharms, line 65: Do I understand correctly that we will start at most `#shards` goroutines? Should we limit how many concurrent goroutines we can start?

dkharms (approved):

query/exec/aggregator.go
forshev: that's correct. we will have to introduce some `column` and `schema` objects

